PR #4, opened 2024-08-29
min amount per deposit

adam-hanna:

Cherry picking from https://github.com/informalsystems/cosmos-sdk/tree/masa/backport-18146-min-amount-per-deposit (related [PR](https://github.com/cosmos/cosmos-sdk/pull/19312/files))


adam-hanna:
There are ALOT of linting issues, which I will slowly chip away at

giunatale:
I think the entire Cosmos SDK codebase was merged here by mistake, while the idea was to cherry pick the x/gov modifications. A hint is the over 2000 files part of the PR most of which are additions. Just pointing it out fyi before it becomes even harder to revert without losing work done

adam-hanna:
> I think the entire Cosmos SDK codebase was merged here by mistake, while the idea was to cherry pick the x/gov modifications. A hint is the over 2000 files part of the PR most of which are additions. Just pointing it out fyi before it becomes even harder to revert without losing work done.
> 
> Should also rebase onto main since we switched over having x/gov own types

This was the minimal import of the cosmos-sdk that it took to cherry pick that commit. I know it's hard to believe but I started with the commit and simply kept importing packages/modules until I could build

giunatale:
> This was the minimal import of the cosmos-sdk that it took to cherry pick that commit. I know it's hard to believe but I started with the commit and simply kept importing packages/modules until I could build

If you look at the PR https://github.com/cosmos/cosmos-sdk/pull/19312/files this is what it touches 
image
and in reality the changes in proto for staking and the change in bank can be discarded, so the PR is focused on x/gov. I don't think you really have to merge the whole sdk into the atomone app. Only the x/gov forked module needs to be updated (and the e2e tests I believe)

PR #5, opened 2024-08-30
feat(gov): disable validator vote inheritance

tbruyelle:


x/gov/keeper/tally.go
```diff
@@ -73,6 +73,7 @@ func (keeper Keeper) Tally(ctx sdk.Context, proposal v1.Proposal) (passes bool,
 		return false
 	})
 
+	/* DISABLED on AtomOne - Voting can only be done with your own stake
```
adam-hanna, line 76: Line 53, above states:
```
// There is no need to handle the special case that validator address equal to voter address.
// Because voter's voting power will tally again even if there will be deduction of voter's voting power from validator.
```

Does this special case need to be handled now that we have disabled Tally?

tbruyelle: Good point, I remember I went to this comment and had the same question. Don't hesitate to refactor now that we have all the unit tests, you can push directly on this branch.

adam-hanna: Ok, I think I correctly added a test for this case:
https://github.com/atomone-hub/atomone/pull/6

If you think it's helpful, feel free to merge into here. If not no worries, this change is good to go!

adam-hanna (approved):

x/gov/keeper/tally.go
tbruyelle: I think this case is already handled, because validators are all setup with a self delegation of 1.
It's handled [here](https://github.com/atomone-hub/atomone/blob/5190b4b0e83b7ba5e5166a1fcf953af5eb5eb79c/x/gov/keeper/tally_test.go#L141) for example: there's only one vote from a validator, so voter address == validator address.

tbruyelle: @adam-hanna actually we can get ride of this comment and simplify a bit the tally function. See #7

x/gov/keeper/tally_test.go
```diff
@@ -213,18 +213,18 @@ func TestTally(t *testing.T) {
 			},
 		},
 		{
-			name: "validator votes yes, inherit delegations",
+			name: "validator votes yes, don't inherit delegations",
```
giunatale, line 216:
```suggestion
			name: "validator votes yes, doesn't inherit delegations",
```

giunatale (changes requested):

x/gov/keeper/tally_test.go
giunatale: very minor change proposed, but otherwise ltgm

giunatale (approved):

PR #8, opened 2024-09-11
add gov/types

tbruyelle:
- copy cosmos-sdk gov/types to x/gov
- rename imports
- rename amino and proto path
- copy cosmos-sdk `proto/cosmos/gov` in `proto/atomone/gov`, rename imports and package, regenerate `*.pb.go` files
- add wrapper for legacy proposal handlers (like in govgen)
- add upgrade and param module client for proposal tx (like in govgen)
- add proposal type registration (required currently executed in related module on the legacy gov module)
- add module types registration where modules used to do it for the legacy gov module


giunatale (approved):
LGTM

PR #9, opened 2024-09-12
feat(x/gov): remove No With Veto, and set default pass threshold to 2/3 and quorum to 25%

giunatale:
- Removes NWV Option
- Sets default pass threshold to 2/3 (more precisely 0.667)
- Sets default quorum to 25%

Changing the default params was pretty straightforward so I did it alongside the NWV removal


tbruyelle (changes requested):
Looks good, but there's a couple of "no_with_veto" strings that still live in other places than `v1beta1`, for instance in the client package.
Search for that specific string and you will find them.

giunatale:
I did leave no with veto in v1beta1

tbruyelle (approved):
LGTM

PR #11, opened 2024-09-12
feat(x/gov): add constitution amendment and law proposals with specific quorum and pass thresholds

giunatale:
The proposal types only make sense in `gov.v1beta1` since `v1` changes to a system where proposals are `sdk.Msg`s. This required therefore touching both packages, and the tally has to check for `v1.MsgExecLegacyContent` because that's the wrapper for legacy proposals (`v1beta1`) in to become an `sdk.Msg` and therefore be stored as a valid `v1.Proposal`.

My initial tests seem to show this works, but will refine further probably with more tests and stuff before promoting it to a full PR.

### DIFF
--- a/x/gov/keeper/tally.go
+++ b/x/gov/keeper/tally.go
@@ -138,7 +138,6 @@ func (keeper Keeper) Tally(ctx sdk.Context, proposal v1.Proposal) (passes bool, 
 	}
 
 	// If more than 2/3 of non-abstaining voters vote Yes, proposal passes
-	threshold, _ := sdk.NewDecFromStr(params.Threshold)
 	if results[v1.OptionYes].Quo(totalVotingPower.Sub(results[v1.OptionAbstain])).GT(threshold) {
 		return true, false, tallyResults
 	}
--- a/tests/e2e/genesis.go
+++ b/tests/e2e/genesis.go
@@ -137,7 +137,7 @@ func modifyGenesis(path, moniker, amountStr string, addrAll []sdk.AccAddress, de
 			sdk.NewCoins(sdk.NewCoin(denom, amnt)), maxDepositPeriod,
 			votingPeriod,
 			quorum.String(), threshold.String(),
-      amendmentsQuorum.String(), amendmentsThreshold.String(), lawQuorum.String(), lawThreshold.String(),
+			amendmentsQuorum.String(), amendmentsThreshold.String(), lawQuorum.String(), lawThreshold.String(),
 			sdk.ZeroDec().String(),
 			false, false, govv1.DefaultMinDepositRatio.String(),
 		),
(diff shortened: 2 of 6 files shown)


giunatale:
I think I can promote this for review. Seems to work fine. For now Laws are basically the exact same as Amendments, but going forward they will have other specific "side effects" in the code so it is nice to separate them already IMV, but let me know, as yes they are just a duplicate for the time being.

I don't like that we have to resort to "legacy" systems but the problem is IMHO with the "new" stuff. On the Cosmos Hub governance still mostly relies on the legacy proposals instead of the v1 version with the arbitrary sdk.Msg. 
We will want to have different proposal types going forward also for other things I suppose so I figure we may consider some alternative path for the atomone x/gov module when we change this.
For now, following the "minimal changes to the fork" idea, I think it's fine to go this way.

x/gov/keeper/tally.go
```diff
@@ -86,6 +87,48 @@ func (keeper Keeper) Tally(ctx sdk.Context, proposal v1.Proposal) (passes bool,
 	// If there is not enough quorum of votes, the proposal fails
 	percentVoting := totalVotingPower.Quo(sdk.NewDecFromInt(totalBondedTokens))
 	quorum, _ := sdk.NewDecFromStr(params.Quorum)
+	threshold, _ := sdk.NewDecFromStr(params.Threshold)
+
+	// Check if a proposal message is an ExecLegacyContent message
+	if len(proposal.Messages) > 0 {
+		var sdkMsg sdk.Msg
+		for _, msg := range proposal.Messages {
```
tbruyelle, line 95: Is there any way to ensure this kind of proposals (constitutional amendment or law) contains only one message ?

Because if I can submit a proposal that contains 2 `MsgExecLegacyContent`, where the first is a constitution amendment and the second a law proposal, the threshold and the quorum will be set to the ones of the laws and not the constituion amendment. It's not a problem when they are equal, but if not it will become a bug.

tbruyelle (commented):
LGTM.

I'm a bit annoyed by relying on v1beta1 for that feature, I would rather rely on the Msg types. Such messages don't exist, but for the constitutional amendment, a new message could make sense (like `MsgUpdateConstitution`) because it would update the Constitution in the store when the proposal passes.
For the law on the other hand, a message like `MsgUpdateLaws` would have nothing to do so it's probably a bit silly.

Just sharing my thoughs, at the end I think your implementation is the best for what we have.

tbruyelle (approved):

x/gov/keeper/tally.go
giunatale: see ValidateBasic. In principle it should, this is just to be sure it's compatible. Also the check on the number of messages is due to how tests were done (the tally bundles the proposal and a bank send) and in reality a proposal in v1 from what I can tell can be an arbitrary number of messages.

giunatale: the threshold is always the highest and law can be lower than regular and constitution can be lower then law. If in the prop there is a constitution amendment then that's what should rule above anything

tbruyelle: oh well yes I forgot about the check "greater than" in the code. So far so good.

PR #13, opened 2024-09-12
feat(x/gov): min amount per deposit

adam-hanna:


adam-hanna:
Ok, the problem with my other PR, is that I tried to bring in every file that was changed by that PR.

Now that I have a better understanding of the cosmos-sdk code base, I know that some of the changes aren't relevant for what we're trying to do now.

This is a much smaller PR :)

tbruyelle (changes requested):
Nice job! It looks pretty good now.
Just one tiny thing, it seems that you (accidentaly?) run gofumpt on the generated proto files. You can see that by re-generate them using `make proto-gen`, the new files are different where they shouldn't. Just fix that and we should be good to merge.

adam-hanna:
> Nice job! It looks pretty good now. Just one tiny thing, it seems that you (accidentaly?) run gofumpt on the generated proto files. You can see that by re-generate them using `make proto-gen`, the new files are differents whereas they shouldn't. Just fix that and we should be good to merge.

Ok, I re-generated them. It seems like there are still some minor formatting differences? Could be because I'm on a diff version of buf?

```bash
$ buf --version
1.32.1
```

tbruyelle:
> Ok, I re-generated them. It seems like there are still some minor formatting differences? Could be because I'm on a diff version of buf?

I don't see any differences, I pulled your changes and ran `make proto-gen`, no changes so all good.

tbruyelle (approved):

PR #20, opened 2024-09-23
chore(x/gov): move law and amendment proposals to the new `v1` system that uses `sdk.Msg`

giunatale:
Changes the law and constitution amendments from being a copy of the legacy text proposal to being compliant with the `v1` system based on `sdk.Msg`s. 
The keeper handlers for now do a no-op in both cases.

Since there is no stateful validation possible, the best thing to do for now is to have the law and amendment proposals be empty and let the content be purely in the summary, i.e. same as "text proposals" for gov/v1.


tbruyelle (approved):
LGTM

PR #22, opened 2024-09-23
chore: test enable cgo

tbruyelle:


tbruyelle:
I can't reproduce the exact signature with binaries built with CGO_ENABLED=1 between GH action and my machine. Closing the PR.

PR #24, opened 2024-09-23
chore: make build-ledger, fix reproducible builds, reinforce go version

tbruyelle:
- add make build-ledger
- fix CI env set TM_VERSION
- reinforce go version to go1.21.13
  - build will exit if go version doesn't match exactly
  - print instruction to run with specific go version
  - add `make print_required_go_version` so CI can fetch the information

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,7 +21,7 @@ jobs:
 
       - uses: actions/setup-go@v5
         with:
-          go-version: "1.20"
+          go-version: "1.21"
 
       - name: Set Env
         run: echo "TM_VERSION=$(go list -m github.com/tendermint/tendermint | sed 's:.* ::')" >> $GITHUB_ENV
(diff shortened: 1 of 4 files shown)


giunatale (approved):

giunatale (approved):
we are enforcing even subversions while technically that would not be needed if not for reproducible builds.

I am not against it, but it's a little tighter as a requirement than what we actually need.

tbruyelle:
> we are enforcing even the subversion while technically that would not be needed if not for consistency of hashes for reproducible builds.
> 
> I am not against it, but it's a little tighter as a requirement than what we actually need.

True, but I have found this to be the easiest way to ensure reproducible builds for most users.

If we remove the subversion check, and just check the major version, how do we know about reproducible builds? We could print a warning, but that would be a 4 line warning with all the instructions, it looks a bit annoying to have that for every build, so I chose the current solution.

Let me know if you think of another way.

tbruyelle:
@giunatale see [d13bc86](https://github.com/atomone-hub/atomone/pull/24/commits/d13bc863075680432341607fb1af7a7b1d845684)

giunatale (approved):

PR #25, opened 2024-09-23
feat(x/gov): constitution amendments can update the constitution

giunatale:
The `ProposeConstitutionAmendment` msg that is used for constitution amendment proposals now contains an `amendment` field, that needs to be in *unified diff* format (the output of `diff -u`) and will trigger an update
of the constitution.

The choice of using a unified diff format instead of requiring to overwrite the constitution with an entirely new string is for a couple reasons:

- human readability
- compactness

The choice of the unified diff format was deemed the best tradeoff to achieve both results.

A cli command in the txs called `generate-constitution-amendment` is also provided that fetches the current constitution and generates the amendment message that needs to go in the proposal, but the patching mechanism accepts standard unified diffs generated from any source. Putting it into the txs seemed the best place.

The code that checks and applies the diffs does not rely on any external library, conversely the CLI utility to generate the diff does make use of `github.com/hexops/gotextdiff`.


tests/e2e/e2e_gov_test.go
```diff
@@ -176,6 +177,35 @@ func (s *IntegrationTestSuite) testGovParamChange() {
...
 
+func (s *IntegrationTestSuite) testGovConstitutionAmendment() {
+	s.Run("constitution amendment", func() {
+		chainAAPIEndpoint := fmt.Sprintf("http://%s", s.valResources[s.chainA.id][0].GetHostPort("1317/tcp"))
+		senderAddress, _ := s.chainA.validators[0].keyInfo.GetAddress()
+		sender := senderAddress.String()
+
+		res := s.queryConstitution(chainAAPIEndpoint)
+		newConstitution := "New test constitution"
+		amendment, err := govtypes.GenerateUnifiedDiff(res.Constitution, newConstitution)
```
tbruyelle, line 188: Even though it is already unit-tested, for end2end completeness it would be nice to use the CLI to generate the proposal file instead, wdyt?

giunatale: sounds good!

x/gov/keeper/constitution_test.go
```diff
@@ -0,0 +1,39 @@
...
+		amendment           string
+		expectedResult      string
+		expectError         bool
+	}{
+		{
+			name:                "failed patch application",
+			initialConstitution: "Hello World",
+			amendment:           "Hi World",
+			expectError:         true,
+		},
```
tbruyelle, line 24: no "success" test case ?

giunatale: Thanks, I actually left this file to readdress later and then forgot about it..

tests/e2e/e2e_gov_test.go
giunatale: done

x/gov/keeper/constitution_test.go
giunatale: done

x/gov/client/cli/tx.go
```diff
@@ -430,7 +436,12 @@ $ %s tx gov generate-constitution-amendment path/to/updated/constitution.md
 		},
 	}
 
-	flags.AddTxFlagsToCmd(cmd)
+	// This is not a tx command (but a utility for the proposal tx), so we don't need to add tx flags.
+	// It might actually be confusing, so we just add the query flags.
+	flags.AddQueryFlagsToCmd(cmd)
```
tbruyelle, line 440: `NewCmdDraftProposal()` is also part of the `tx` command and just output some data, similarly. But it doesn't replace the tx flags with the query flags. Why do you need to query flags ?

giunatale: `NewCmdDraftProposal` is a little different also because it's generating a tx right? while this command doesn't, it generates the content for a tx NOT the whole tx so stuff like chain-id or fees doesn't even make sense.
While it could make sense in that case.

But, it's true that adding the tx flags to that command is alwo wrong because there are some flags that let imagine that the command is able to submit the tx too

giunatale: After further checks it seems actually the txflags are added but ignored in the case of `NewCmdDraftProposal`? So shouldn't they be removed altogether? no flags are used aside from the skip-metadata one (generally the flags are used in `tx.GenerateOrBroadcastTxCLI()` but that is not called in the command as the command writes directly to file, actually the name of the file can't even be changed...)

giunatale: Confirmed, so I actually removed it in this PR

tbruyelle: got it, thanks for the investigation and the report!

tbruyelle (approved):

PR #27, opened 2024-09-25
doc: gov doc revamp

tbruyelle:
TODO:
- [x] include late quorum voting period extension
- [x] include constitution amendments and law proposals
- [x] add new params
- [x] check other md files


giunatale:
We also need to add constitution amendments

giunatale:
@tbruyelle this should be done, please review

tbruyelle:
@giunatale LGTM

Issue #35, opened 2024-10-22
Fix swagger generation


tbruyelle:
The swagger files should be generated by the `make proto-swagger-gen` command, but it fails:

```
$ make proto-swagger-gen
Generating Protobuf Swagger
Failure: could not read file buf.gen.swagger.yaml: open buf.gen.swagger.yaml: no such file or directory
make: *** [Makefile:338: proto-swagger-gen] Error 1
```

The same error is triggered with Gaia v15.2.0 where this repo is forked from.

PR #38, opened 2024-10-25
fix: swagger generation

tbruyelle:
Closes: #35

The swagger generation consists of the following steps:
- generate json files from proto using the swagger plugin
- combine json files into client/docs/swagger-ui/swagger.yaml file
- embed client/docs/swagger-ui into go code using rakyll/statik
- serve this specific statik fs to `/swagger` endpoint


giunatale:
there is a lint error, the e2e error is instead I think a false positive

giunatale (approved):

PR #39, opened 2024-10-28
chore: add info log when building with Ledger support disabled

giunatale:
Add an info log line when building without ledger support to inform the user of the presence of the appropriate target / env variable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,11 +104,16 @@ ifneq ($(GO_SYSTEM_VERSION), $(GO_REQUIRED_VERSION))
 	exit 1
 endif
 
+check_ledger:
+ifeq ($(LEDGER_ENABLED),false)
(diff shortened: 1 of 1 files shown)


tbruyelle (approved):

Makefile
```diff
@@ -104,11 +104,16 @@ ifneq ($(GO_SYSTEM_VERSION), $(GO_REQUIRED_VERSION))
 	exit 1
 endif
 
+check_ledger:
+ifeq ($(LEDGER_ENABLED),false)
+	$(info Building without Ledger support. Set LEDGER_ENABLED=true to enable or use build-ledger target to build with Ledger support.)
```
tbruyelle, line 109:
```suggestion
	$(info Building without Ledger support. Set LEDGER_ENABLED=true or use build-ledger target to build with Ledger support.)
```

tbruyelle (changes requested):

PR #41, opened 2024-11-02
bump cometbft to avoid serious issues

faddat:


tbruyelle:
Hey thank you for this PR.

I notice that the required go version is updated from 1.21 to 1.22 (probably because one of the dependencies required by `cometbft@v0.37.13` declares this version), and if I remember correctly, the reason why the go version is fixed in SDK chain binaries is because there are some sync issues when the binary is built with different go versions.

So my first concern about your change is that we should carefully upgrade the go version, probably better for a major next release, to ensure that all validators are using a binary built with the same go version.

WDYT?

giunatale:
I also see some tests failing that need to be addressed.

Besides, can you articulate exactly why we would need to bump these dependencies? The way it's put up for now is not really substantiated. We would need precise reasons to upgrade dependencies.

Issue #44, opened 2024-11-07
Photon implementation


tbruyelle:
From the AtomOne Constitution [Article 3 Section 5](https://github.com/atomone-hub/genesis/blob/main/CONSTITUTION.md#section-5-the-photon-token)

> Section 5: The PHOTON Token
> The PHOTON shall be the only fee token except for ATONE to PHOTON burn transactions. This applies for all transactions on the root and core shards, and all IBC and ICS payments.
> 
> ATONE tokens may be burned to PHOTON tokens at a conversion rate set by law such that the total amount of PHOTONs mintable through burning ATONE tokens shall be capped at 1B PHOTON tokens.
> 
> PHOTONs cannot be converted back into ATONE tokens.

The implementation work will be splitted in multiple tasks, each possibly involving a different PR:

- [x] The photon ADR (#34) 
- [x] The photon mint message (#53)
- [x] The photon ante decorator (#54)
- [x] The photon denom metadata migration (#55)
- [x] Documentation about validator config and `minimum-gas-price` (#51)

PR #49, opened 2024-11-18
feat(x/photon): validate fee ante decorator

tbruyelle:
Relates to #44 

Warning: because the base branch doesn't exist in the origin repo, this PR couldn't used `tbruyelle/feat/photon-module` as the base branch. The real diff can be seen here : https://github.com/tbruyelle/atomone/compare/tbruyelle:atomone:tbruyelle/feat/photon-module...tbruyelle/feat/photon-tx-fee-checker

This PR adds a `sdk.AnteDecorator` implementation in `x/photon/ante` and adds it to the list of the app ante decorators. This new ante decorator ensures that photon is the fee token or returns an error. Exceptions are possible like if the transaction's messages type URL are present in a new parameter of the photon module `txFeeExceptions`.

Some significant changes have been made to the e2e tests to make them compatible with this conditions.

`uatone` has also been put in a constant to avoid repetition (in `app/params` package)


tbruyelle:
closed in favor of #54

PR #51, opened 2024-11-25
docs: v2 upgrading instructions

tbruyelle:
Relates to #44 

Adapted version of https://github.com/cosmos/gaia/blob/v15.0.0/UPGRADING.md, with new section for the `minimum-gas-prices` change.

[Rendered Version]()


UPGRADING.md
```diff
@@ -1,3 +1,276 @@
...
+
+It is critically important for validator operators to back-up the
+`.atomone/data/priv_validator_state.json` file after stopping the atomoned
+process. This file is updated every block as your validator participates in
+consensus rounds. It is a critical file needed to prevent double-signing, in
+case the upgrade fails and the previous chain needs to be restarted.
+
+### Testing
+
+XXX UPDATE?
```
giunatale, line 78: Let's revisit as we make plans to run a testnet. May be worth removing the text from this section so we can unblock this PR (and update later) or wait until we have a clearer picture

UPGRADING.md
```diff
@@ -1,3 +1,276 @@
...
+
+```shell
+cosmovisor run start --x-crisis-skip-assert-invariants --home $DAEMON_HOME
+```
+
+XXX NOTE: keep that sentence ??
+
+Skipping the invariant checks is strongly encouraged since it decreases the
+upgrade time significantly and since there are some other improvements coming
+to the crisis module in the next release of the Cosmos SDK.
```
giunatale, line 210:
```suggestion
Skipping the invariant checks can be used to decreases the
upgrade time significantly, but it is done at the expense of
verifying state validity and must be done cautiously.
```

giunatale (changes requested):

UPGRADING.md
tbruyelle: Ok for addressing that later and not block this PR, I removed that section [4f21380](https://github.com/atomone-hub/atomone/pull/51/commits/4f213805e5e8f5d8300e8733b2a288450083b7f5)

giunatale (approved):

PR #53, opened 2024-11-26
feat(x/photon): mint photon message

tbruyelle:
Relates to #44

This PR adds a new message `MsgMintPhoton` inside a new `x/photon` module.
List of changes:
- `x/photon` queries : `params` & `conversion_rate`
- `x/photon` transaction: `MsgMintPhoton`
- app plumbing with the new module
- Some relative changes were made to the e2e tests to remove anything related to the `stake` denom (use `uatone` instead), which had some impact on other tests that had to be fixed.
- Makefile:
  - target `start-localnet-ci` use `uatone` denom instead of default `stake`
  - new target `mockgen` to generates the mocks from gov and photon expected keepers


x/photon/keeper/msg_server.go
```diff
@@ -0,0 +1,113 @@
...
+		uphotonToMint   = bondDenomToBurn.Amount.ToLegacyDec().Mul(conversionRate).RoundInt()
+	)
+	// If no photon to mint, do not burn bondDenomToBurn, returns an error
+	if uphotonToMint.IsZero() {
+		return nil, types.ErrZeroMintPhotons
+	}
+	// If photonToMint + photonSupply > photonMaxSupply, returns an error
+	if uphotonSupply.Add(uphotonToMint.ToLegacyDec()).GT(sdk.NewDec(types.MaxSupply)) {
+		return nil, types.ErrNotEnoughPhotons
+	}
```
giunatale, line 54: Both checks but especially the second could be avoided. The second one especially, considering you get the conversionRate and its implicit  photonToMint + photonSupply <= photonMaxSupply because the conversionRate considers mintablePhotons = photonMaxSupply - photonMinted.

Seemengly, for the reasons explained in the ADR if the blockchain is producing blocks it means a non-zero amount of bond denom is staked, meaning it's impossible to convert **all** bond denom in cirulation and therefore the cap will never be reached.

x/photon/keeper/msg_server.go
```diff
@@ -0,0 +1,113 @@
...
+	// Compute photons to mint
+	var (
+		bondDenomSupply = k.bankKeeper.GetSupply(ctx, bondDenom).Amount.ToLegacyDec()
+		uphotonSupply   = k.bankKeeper.GetSupply(ctx, types.Denom).Amount.ToLegacyDec()
+		conversionRate  = k.conversionRate(ctx, bondDenomSupply, uphotonSupply)
+		bondDenomToBurn = msg.Amount
+		uphotonToMint   = bondDenomToBurn.Amount.ToLegacyDec().Mul(conversionRate).RoundInt()
+	)
+	// If no photon to mint, do not burn bondDenomToBurn, returns an error
```
giunatale, line 47:
```suggestion
	// If no photon to mint, do not burn bondDenomToBurn, returns an error
	// this could happen due to rounding
```

x/photon/keeper/msg_server.go
```diff
@@ -0,0 +1,113 @@
...
+	params := k.GetParams(ctx)
+	if params.MintDisabled {
+		return nil, types.ErrMintDisabled
+	}
+
+	// Ensure burned amount denom is bond denom
+	bondDenom := k.stakingKeeper.BondDenom(ctx)
+	if msg.Amount.Denom != bondDenom {
+		return nil, types.ErrBurnInvalidDenom
+	}
```
giunatale, line 38: check also that there is a non-zero amount for the denom coin

x/photon/keeper/msg_server.go
giunatale: The first error can indeed happen due to rounding issue so it should stay, but the second check can be safely removed as agreed.

tbruyelle: I removed the second check [3c9199f](https://github.com/atomone-hub/atomone/pull/53/commits/3c9199fd5e9e9a473e5121cf02e6e21753b946d3).
As you noted the first one is required when the photon to mint becomes 0 after rounding.

x/photon/keeper/msg_server.go
giunatale: I would have still added the check here too but ok, I guess we'll check with a live test if this is a problem or not. I think you are right but let's see. Closing for now

giunatale (approved):

PR #54, opened 2024-11-26
feat(x/photon): validate fee ante decorator

tbruyelle:
Relates to #44 

This PR adds a `sdk.AnteDecorator` implementation in `x/photon/ante` and adds it to the list of the app ante decorators. This new ante decorator ensures that photon is the fee token or returns an error. Exceptions are possible like if the transaction's messages type URL are present in a new parameter of the photon module `txFeeExceptions`.

Some significant changes have been made to the e2e tests to make them compatible with this conditions.

`uatone` has also been put in a constant to avoid repetition (in `app/params` package)


x/photon/ante/ante.go
```diff
@@ -0,0 +1,77 @@
...
+		// feeDenom not allowed
+		return ctx, sdkerrors.Wrapf(types.ErrInvalidFeeToken, "fee denom %s not allowed", feeDenom) //nolint:staticcheck
+	}
+	// feeDenom photon is allowed
+	return next(ctx, tx, simulate)
+}
+
+// isTxFeeExcepted returns true if all tx messages type URL are presents in
+// txFeeExceptions, or if it starts with a wildcard "*".
+func isTxFeeExcepted(tx sdk.Tx, txFeeExceptions []string) bool {
```
giunatale, line 61:
```suggestion
func allowsAnyTxFee(tx sdk.Tx, txFeeExceptions []string) bool {
```

giunatale (changes requested):
Quite more extensive than what initially seemed to be. Mostly lots of changes required for tets and the issue with simulation.

One note, "Excepted" is rarely used and rarely I see it, I would use something else. I added a suggestion to the function name and I would also do this everywhere "excepted" is used, e.g. in the proto

tbruyelle:
> Quite more extensive than what initially seemed to be. Mostly lots of changes required for tests and the issue with simulation.
> 
> One note, "Excepted" is rarely used and rarely I see it, I would use something else. I added a suggestion to the function name and I would also do this everywhere "excepted" is used, e.g. in the proto

Removed the 'excepted' term and renamed the function as suggested 15374464

x/photon/ante/ante.go
tbruyelle: done 15374464

giunatale (approved):

PR #56, opened 2024-11-28
docs(upgrade): add relayer config change section

tbruyelle:
Relates to #44 

I realized that the running relayers would also have to update their config to use PHOTON for the fees, or else all txs would be rejected. So added a section to cover that part.

Also added a kind remind for validators that have `authz` transactions submitted periodically (for example validators who support services like restake) to update the tx fee denom as well.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -8,7 +8,7 @@ upgrade successfully for the AtomOne v2 release.
 For more details on the release, please see the [release notes][v2].
 
 **Validators** will have to change their configuration to allow the PHOTON
-denom for the fees, **before** the upgrade, see [Validator config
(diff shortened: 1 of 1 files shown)


UPGRADING.md
```diff
@@ -7,6 +7,14 @@ upgrade successfully for the AtomOne v2 release.
 
 For more details on the release, please see the [release notes][v2].
 
+**Validators** will have to change their configuration to allow the PHOTON
+denom for the fees, **before** the upgrade, see [Validator config
```
giunatale, line 11:
```suggestion
denom (`uphoton`) for the fees, **before** the upgrade, see [Validator config
```

giunatale (approved):

PR #65, opened 2024-12-17
docs: ADR-003 Governance Proposal Deposit Auto-Throttler

giunatale:
ADR for the implementation of the min deposit throttling, with notes on also the min initial deposit.

Implementation is at https://github.com/atomone-hub/atomone/pull/69

[Rendered version](https://github.com/atomone-hub/atomone/tree/docs/adr-dyamic-deposit/docs/architecture/adr-003-governance-proposal-deposit-auto-throttler.md)

### DIFF
--- a/docs/architecture/adr-003-governance-proposal-deposit-auto-throttler.md
+++ b/docs/architecture/adr-003-governance-proposal-deposit-auto-throttler.md
@@ -66,22 +66,30 @@ An update of the `MinDeposit` can be triggered **by the passage of a certain amo
 
 $$
 D_{t+1} = \max(D_{\min}, D_t \times (1+ sign(n_t - N) \times \alpha \times \sqrt[k]{| n_t - N |}))
-\\
+$$
+
+$$
 \alpha = \begin{cases} \alpha_{up} & n_t \gt N \\
 \alpha_{down} & n_t \leq N
 \end{cases}
-\\
+$$
+
+$$
 sign(n_t - N) = \begin{cases} 1 & n_t \geq N \\
 -1 & n_t \lt N
-\end{cases} \\ k \in \N \\
-0 \lt \alpha_{down} \lt 1 \\ \\ 0 \lt \alpha_{up} \lt 1 \\ \alpha_{down} \lt \alpha_{up}\\ 
+\end{cases}
 $$
+
+$$k \in {1, 2, 3, ...}$$
+$$0 \lt \alpha_{down} \lt 1$$
+$$0 \lt \alpha_{up} \lt 1$$
+$$\alpha_{down} \lt \alpha_{up}$$
 
 Where:
 
 - $D_{t+1}$ is the new deposit value, $D_{\min}$ is the floor deposit value and $D_t$ the deposit at time $t$
 - $n_t$ is the number of active proposals at time $t$ and $N$ the target number of proposals
-- $k$ expresses the *sensitivity* of the rate of increase or decrease to the distance of the number of active proposals $n_t$ with respect to the target $N$. Since the rate of change is proportional to this distance, $k$ can be used to tune how much getting further away from $N$ exacerbates things.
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-003-governance-proposal-deposit-auto-throttler.md
```diff
@@ -0,0 +1,161 @@
...
+
+While the ADR is focused on presenting the dynamic adjustment of `MinDeposit`, the same mechanism can be used to dynamically adjust `MinInitialDeposit` - with different values for parameters and focusing on proposals entering or exiting the deposit period instead of the voting period. The two dynamic systems would function in the same way, but independently of each other, allowing even `MinInitialDeposit` to grow bigger than `MinDeposit` at times.
+
+The proposed mechanism can be vaguely compared to the auto-adjusting inflation rate that targets a 2/3 bonding ratio. However, in this case, it is used to automatically adjust the deposit based on the number of proposals.
+
+## **Context[](https://docs.cosmos.network/main/architecture/adr-template#context)**
+
+At the time of writing, the `x/gov` module on AtomOne uses a `MinDeposit` ****parameter set to 512 ATONEs, while the `MinInitialDepositRatio` is equal to 10% of the `MinDeposit`, which makes the `MinInitialDeposit` to be 51.2 ATONEs.
```
tbruyelle, line 26:
```suggestion
At the time of writing, the `x/gov` module on AtomOne uses a `MinDeposit` parameter set to 512 ATONEs, while the `MinInitialDepositRatio` is equal to 10% of the `MinDeposit`, which makes the `MinInitialDeposit` to be 51.2 ATONEs.
```

docs/architecture/adr-003-governance-proposal-deposit-auto-throttler.md
```diff
@@ -0,0 +1,161 @@
...
+### Implementation
+
+The following parameters are added to the `x/gov` module params:
+
+- `min_deposit_floor`: floor value for the minimum deposit required for a proposal to enter the voting period
+- `min_deposit_update_period`: duration that dictates after how long the dynamic minimum deposit should be recalculated for time-based updates.
+- `target_active_proposals`:  the number of active proposals the dynamic minimum deposit should target.
+- `min_deposit_increase_ratio`: the ratio of increase for the minimum deposit when the number of active proposals exceeds the target by 1.
+- `min_deposit_decrease_ratio`: the ratio of increase for the minimum deposit when the number of active proposals is 1 less than the target.
+- `min_deposit_sensitivity_target_distance`: A positive integer representing the sensitivity of the dynamic minimum deposit increase/decrease to the distance from the target number of active proposals. The higher the number, the lower the sensitivity. A value of 1 represents the highest sensitivity.
```
tbruyelle, line 122: I would advocate for having a single struct handling all these related parameters. It makes even more sense considering we would have to duplicate all of them for the `MinInitialDeposit`.

giunatale: A separate struct for params is a deprecated method (see the other *Params structs) so that's why I didn't do it.

tbruyelle: I don't think the reason for the deprecation of the other `*Params` structs was related to the way the global `Params` struct should be defined.

In the [ADR](https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-046-module-params.md) that deprecates x/params, it is clearly stated that :
> Developers can choose to either create a unique data structure for every field in Params or they can create a single Params structure as outlined above in the case of x/auth.

Looking at the [PR](https://github.com/cosmos/cosmos-sdk/pull/12631/files#diff-6459d5ff0e84a4b446613a13ddb888cceeaa22e3cf34eb6398c472b448cfc835) that deprecated x/params for the gov module, it is more likely that these structs were removed because they were not part of a global struct that could be used as the global `Params`. In fact, they were part of the `GenesisState` struct, so it was necessary to merge them into a new global `Params` struct.

For example, the `SendEnabled` field of the bank module param is a custom struct.

giunatale: Ok, this sounds good. I made the change on both code and adr

PR #73, opened 2025-01-14
feat(x/gov): add governors

giunatale:
iterating over the work started in #16, this PR adds governors to the x/gov module.

More details can be also read in #16


giunatale:
since this work was left untouched for a while, we need to give it a final pass to identify potentially missing tests and check test coverage, and add documentation. Ideally we would also want to do some more in depth profiling/benchmarking to assess the performance impact of this addition.

proto/atomone/gov/v1/query.proto
```diff
@@ -61,6 +61,31 @@ service Query {
...
+  }
+
+  // GovernanceDelegations queries all delegations of a governor.
+  rpc GovernanceDelegations(QueryGovernanceDelegationsRequest) returns (QueryGovernanceDelegationsResponse) {
+    option (google.api.http).get = "/atomone/gov/v1/governors/{governor_address}/delegations";
+  }
+
+  // GovernanceDelegation queries a delegation
+  rpc GovernanceDelegation(QueryGovernanceDelegationRequest) returns (QueryGovernanceDelegationResponse) {
+    option (google.api.http).get = "/atomone/gov/v1/delegations/{delegator_address}";
```
tbruyelle, line 82: I would rather see singular `delegation` here, as it will always return a single.

giunatale: This is for consistency, the plural is used in other places too (e.g. proposals)

tbruyelle: Understandable, then the first one `/atomone/gov/v1/governor/{governor_address}` should be `/atomone/gov/v1/governors/{governor_address}` ?

giunatale: > Understandable, then the first one `/atomone/gov/v1/governor/{governor_address}` should be `/atomone/gov/v1/governors/{governor_address}` ?

yes

x/gov/types/address.go
julienrbrt, line 1: Addresses should be derived using https://github.com/atomone-hub/cosmos-sdk/blob/f5e0eeff6321f4a2a399aaffd0b149c51a892e72/x/auth/types/credentials.go#L36

x/gov/keeper/delegation.go
julienrbrt, line 1: maybe a good exercice to use collections here?

julienrbrt (commented):
preliminary review, i haven't check the whole business logic yet.

docs/architecture/adr-006-governors.md
```diff
@@ -0,0 +1,121 @@
...
+
+1. Introduction of `Governor` Entities:
+   - A `Governor` is created through a `MsgCreateGovernor` transaction.  
+   - The system enforces a minimum governance self-delegation requirement
+   before an account can achieve active governor status, which in practice
+   translates to a minimum stake requirement.
+   - The system tracks each governor’s address, description, status (e.g., active),
+   and last status change time (to limit frequent status toggles).
+
+2. Delegation and Undelegation:
```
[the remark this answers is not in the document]
giunatale: Yes, at proposal tally.

Also, the limitations to `x/gov` for the canonical module are indeed inherited from the dependency with `x/staking` that we don't have here. So indeed there is no need to have these limitations in place. The representative voting system can be more "liquid" as a result.

docs/architecture/adr-006-governors.md
```diff
@@ -0,0 +1,121 @@
...
+   from the governor’s aggregated shares to avoid double counting (similarly
+   to how it's done in canonical Cosmos-SDK governance for validators).
+
+4. Transitioning Governor Status:
+   - A governor can switch their status (e.g., from inactive to active) by meeting
+   the min self-delegation threshold and updating their status with a
+   `MsgUpdateGovernorStatus`. Governors that set themselves to inactive are allowed
+   to delegate their governance voting power to another governor. A status change
+   can however only occur after a waiting period (`GovernorStatusChangePeriod`).
+   - Attempts to rapidly change status are disallowed; changes can only occur
```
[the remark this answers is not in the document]
giunatale: So you are proposing to prevent a status change if there is at least a proposal active (i.e. in voting period?)

Anyway, it would be the same as validators, that can unbond regardless of the status of proposals. We can further discuss this, but on a first level, what is your concern?

x/gov/keeper/delegation.go
julienrbrt: less boilerplate basically, you define the collections in the keeper and you do not need this whole boilerplate of getters and setters.

proto/atomone/gov/v1/gov.proto
```diff
@@ -367,3 +375,77 @@ message QuorumRange {
...
+// validator's bond shares for a set amount of validators. When a delegator
+// delegates a percentage of its x/gov power to a governor, the resulting
+// shares from each delegators delegations in x/staking are added to the
+// governor's total shares.
+message Governor {
+  option (gogoproto.equal)           = false;
+  option (gogoproto.goproto_getters) = false;
+
+  // governor_address defines the address of the governor; bech32-encoded.
+  string governor_address = 1;
```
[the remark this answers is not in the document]
julienrbrt: mmh, given this is a derived address, maybe it isn't not needed indeed.

x/gov/keeper/delegation.go
tbruyelle: I love collections, it is great piece of work. It has just the required abstraction to make dev life easier w/o losing control. That said, to use it for governor we need to wait for SDK50 merge (or to use the [port](https://github.com/tbruyelle/atomone/tree/tbruyelle/feat/multisig/collections) I made for onchain multisig, but that's not reasonnsable since sdk50 upgrade is close  ^^)

docs/architecture/adr-006-governors.md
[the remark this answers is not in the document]
giunatale: yeah, @tbruyelle brings up a great point. I think it's ok to leave the system as is for this reason. Moreover, even if we wanted to change it (something which I actually worked on), we would have the issue of having to also prevent staking undelegations which bring the bonded amount below the min stake requirement, otherwise that would also be a way to bypass the queued change. And in practice the hooks do not allow to distinguish between an undelegation and a redelegation, which make doing this very challenging.

For the reason above and the observation from Thomas, I think it's better if we keep this as is

giunatale:
> im realizing now that will have to come too to my gigantic PR 😅

I will deal with it, once the gov module update is done I will adapt this PR to port the governors.
It will definitely will be painful, but it is what it is

PR #77, opened 2025-02-03
deps: upgrade cometbft to v0.37.15

tbruyelle:
Fixes 2 cometbft securities issues (ASA-2025-001, ASA-2025-002).

Upgrade of go is required because upgraded deps use a later version.

Targets the `release/v1.x` branch which is based on `v1.0.0` plus the change related to the max voting period adjustable via ldflags.


giunatale:
Do we know if this can be a minor release and if it affects consensus - i.e. requiring a chain upgare?
Otherwise, we should issue a new minor release and warn nodes to upgrade

Also, we need to port this over main as well I guess

tbruyelle:
I went through the cometbft CL between v0.37.5 and v0.37.15 and as expected there is no consensus breaking change (I assume that such change would not go into a patch release, but at least a minor release).
About the upgrade in main, it os here #78 .

PR #81, opened 2025-02-10
chore: refactoring docker system

albttx:


Dockerfile
```diff
@@ -1,20 +1,22 @@
-ARG IMG_TAG=latest
+ARG GO_VERSION="1.22.10"
```
tbruyelle, line 1: Let's not have a default value here. I assume without it the gh action will fail because it currently relies on it, but this is not what we want. We want the GO_VERSION to match the version declared in the go.mod, like in `make docker-build-debug`. 

Probably the simplest way to handle this is to call `make docker-build-debug` in the gh action (or maybe use an alias `make docker-build`).

tbruyelle (changes requested):

Dockerfile
albttx: Mhhh, i don't like not having value here, because a simple `docker build -t my-img .` would fail

albttx: Or we can move the `-alpine` in the makefile, and have by default `alpine` which will take the latest go version ?

tbruyelle: Yes, and it's not a problem per say.

The problem here is if we upgrade the go version in the go.mod, the docker build will fail because the `make install` command will return an error because the go versions (docker image and go.mod) does not match. 

In other words, if we upgrade the go version in the go.mod, we also have to change this `ARG` here, and this is what I want to avoid.

Maybe it is possible to guess the go version in a preliminary stage of the docker file, where `go list -f {{.GoVersion}} -m` is executed just like in the Makefile. If yes then we don't need that `ARG`. But I'm not sure it is possible to define `ARG` with executed content inside the docker file.

tbruyelle: > Or we can move the `-alpine` in the makefile, and have by default `alpine` which will take the latest go version ?

But the go version of the docker build must match the go of the go.mod, like the restriction we have in the makefile. I don't understand why do you want to ignore the restrictions.

albttx: Ok, i've updated this way :)

tbruyelle: Thank you, but I think the gh action needs to be updated accordingly. The step `docker/build-push-action@v5` will not feed the GO_VERSION arg I suppose. Is there a way to test this gh action?

tbruyelle: @albttx I think my comment still applies. So far I don't see how the gh action passes the `GO_VERSION` to the docker build so I think the gh action will fail. Can you trigger it manually ? Since you pushed from a fork I can't select your branch to run it manually, but maybe there's an other way.

albttx:
https://github.com/albttx/atomone/actions/runs/13855138885/job/38770235609

Working on my fork, i'm good to go

tbruyelle (approved):
All the settings you added to the github actions are a bit cryptic to me, but LGTM anyway.

PR #90, opened 2025-03-03
feat(x/gov): burn deposit if no votes > `BurnDepositNoThreshold`

tbruyelle:
`BurnDepositNoThreshold` has a default value of `0.8`.


giunatale (changes requested):
Just one change I would suggest doing, but aside from that it looks good

x/gov/keeper/tally.go
```diff
@@ -24,30 +24,47 @@ func (keeper Keeper) Tally(ctx sdk.Context, proposal v1.Proposal) (passes bool,
...
+	yesPercent := results[v1.OptionYes].Quo(activeVotingPower)
+	if yesPercent.GT(threshold) {
+		return true, false, tallyResults, nil
 	}
 
-	if results[v1.OptionYes].Quo(totalVotingPower.Sub(results[v1.OptionAbstain])).GT(threshold) {
-		return true, false, tallyResults
+	// If more than `burnDepositNoThreshold` of non-abstaining voters vote No,
+	// proposal is rejected and deposit is burned.
+	burnDepositNoThreshold, err := sdk.NewDecFromStr(params.BurnDepositNoThreshold)
```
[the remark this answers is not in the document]
tbruyelle: absolutely correct [19e693e](https://github.com/atomone-hub/atomone/pull/90/commits/19e693e123171f26b6439dd462f830bc504724b7)

giunatale (approved):
let's hold off from merging for now, but LGTM

giunatale:
actually, looking at this again I think we made the same mistake in `getQuorumAndThreshold` and we can remove the error in `Tally` altogether because `getQuorumAndThreshold` should not return an error and use the Must APIs

tbruyelle:
> actually, looking at this again I think we made the same mistake in `getQuorumAndThreshold` and we can remove the error in `Tally` altogether because `getQuorumAndThreshold` should not return an error and use the Must APIs

Good catch, I converted all to `MustNew*` and so removed the returned error from `getQuorumAndThreshold` and from `Tally` function. [658af5d](https://github.com/atomone-hub/atomone/pull/90/commits/658af5dc4a6045c35606c0166f989203e9593994)

giunatale (approved):
TY!

x/gov/types/v1/params.go
```diff
@@ -419,5 +423,16 @@ func (p Params) ValidateBasic() error {
 		return fmt.Errorf("minimum initial deposit decrease ratio too large: %s", minInitialDepositDecreaseRatio)
 	}
 
+	burnDepositNoThreshold, err := sdk.NewDecFromStr(p.BurnDepositNoThreshold)
```
giunatale, line 426: We need to add a check for threshold, law and amendment threshold

so that the `burnDepositNoThreshold > 1 - threshold` (or the burn would happen with less than no votes to make a proposal fail, which makes no sense since the proposal would mean is passing)

the same check goes for law and amendment thresholds

giunatale (changes requested):

x/gov/README.md
```diff
@@ -199,6 +201,11 @@ The initial option set includes the following options:
 `Abstain` option allows voters to signal that they do not intend to vote in
 favor or against the proposal but accept the result of the vote.
 
+At the end of the voting period, if the percentage of `No` votes (excluding
+`Abstain` votes) is greater than a specific threshold (see [Burnable
+Params](#burnable-params) section), then the proposal is considered as SPAM and
+its deposit is burnt.
```
giunatale, line 207:
```suggestion
its deposit is burned.
```

x/gov/types/v1/params.go
```diff
@@ -419,5 +423,16 @@ func (p Params) ValidateBasic() error {
...
+	if err != nil {
+		return fmt.Errorf("invalid burnDepositNoThreshold string: %w", err)
+	}
+	if !burnDepositNoThreshold.IsPositive() {
```
giunatale, line 430: I would replace this check with the `burnDepositNoThreshold > 1 - threshold` && `burnDepositNoThreshold > 1 - lawThreshold` && && `burnDepositNoThreshold > 1 - amendmentThreshold`

x/gov/types/v1/params.go
tbruyelle: If the proposal is passing, the `burnDepositNoThreshold` is not checked, so the burn would not happen anyway. But I will still do this `burnDepositNoThreshold > 1 - threshold` check so the condition is not tied to the order of code execution.

tbruyelle: done here 3788dbb7

giunatale (approved):

PR #91, opened 2025-03-06
feat(x/feemarket): sync with `skip/feemarket` `v1.1.1` (except tests) and re-add postHandler

giunatale:
as the title suggests.

change to 0 fees in simulation to allow for no bank balances in sim, as per https://github.com/skip-mev/feemarket/pull/138
I also re-added the postHandler since wrongly removed also the logic to updated consumed gas in the module state

missing backporting tests from https://github.com/skip-mev/feemarket/pull/138


giunatale:
tests failing is normal, the branch is still in an inconsistent state, so would not worry if they fail for now, as it's expected

tbruyelle (approved):

PR #93, opened 2025-03-07
chore: remove `x/crisis` module

giuliostramondo:
This PR removes `x/crisis` because it has been identified as a security issue. 

Originally, the module could be used by anyone to halt the chain in the event of a violated invariant. To prevent spamming the requested would be charged a `ConstantFee` .

However, the module has been shown to **NOT** [halt the chain](https://github.com/cosmos/cosmos-sdk/security/advisories/GHSA-qfc5-6r3j-jj22)  and to [not charge](https://github.com/cosmos/cosmos-sdk/security/advisories/GHSA-w5w5-2882-47pc) the `ConstantFee`  to the requester.

This implies that an attacker could have used `x/crisis` as a vector for a DOS attack.


CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 
 - Upgrade CometBFT to v0.37.15 to fix securities issues (ASA-2025-001, ASA-2025-002) [#78](https://github.com/atomone-hub/atomone/pull/78)
 - Upgrade ibc-go to v7.9.2 to fix ASA-2025-004 [#84](https://github.com/atomone-hub/atomone/pull/84) [#85](https://github.com/atomone-hub/atomone/pull/85)
+- Remove x/crisis [#93](https://github.com/atomone-hub/atomone/pull/93)
```
giunatale, line 38:
```suggestion
- Remove `x/crisis` [#93](https://github.com/atomone-hub/atomone/pull/93)
```

tbruyelle (approved):
With @giuliostramondo, we upgraded a local v1 chain with this code using a software upgrade proposal and the upgrade went well. 

Then we wanted to ensure that the `crisis` module was actually removed from the store.

The `atomoned export` command does not output the `crisis` module state, but since the logic of export is based on the module list, it's kind of obvious because the `crisis` module is no longer part of that list. Which means that not having the `crisis` module in the output of the `export` command does not necessarily mean that the db no longer contains information about the module.

So we went a bit further and tried to fetch the crisis module keys from the store, using the appropriate prefix:

```go
	prefix := "s/k:crisis/"
	db = dbm.NewPrefixDB(db, []byte(prefix))
	it, err := db.Iterator(nil, nil)
	if err != nil {
		panic(err)
	}
	for it.Valid() {
		fmt.Println(string(it.Key()))
		it.Next()
	}
```

And it appears that the crisis store is still there after the upgrade, so we need to understand why. 

However, the code in this PR sounds correct to me, so I approve it. Maybe there's a bug in the SDK or something we've misinterpreted.

tbruyelle (changes requested):
So after more tests regarding the deletion of the crisis store, it appears that we need to keep the crisis key in the return of `AppKeepers.GenerateKeys()`, with a comment explaining why it needs to be there while the module is removed, and also saying that this should be removed for the next release.

tbruyelle (approved):

PR #100, opened 2025-03-17
Revamp feemarket tests

tbruyelle:
- Align tests design with the one we have in photon and gov module
- Remove test suite that makes use of `skip-mev/chaintestutil` (and so remove that dep)
- Add more cases for untested scenarios
- Leave the tested code unchanged for now, but some improvements are likely to be possible once the tests have been merged.


giunatale:
for simulation, I think it will be difficult to fix without changing how `GenAndDeliverTxWithRandFees()` in `x/simulation` from the cosmos-sdk works, so that the random fee selected it at least above the minimum.

The other solution would be to disable the fee system during simulation so that it falls back on the original fee deductor. 
The system is enabled by setting the related boolean to true by default in `module.go:DefaultGenesis()` which calls `DefaultGenesisState()` on the non AIMD `params.go`

tbruyelle:
> for simulation, I think it will be difficult to fix without changing how `GenAndDeliverTxWithRandFees()` in `x/simulation` from the cosmos-sdk works, so that the random fee selected it at least above the minimum.
> 
> The other solution would be to disable the fee system during simulation so that it falls back on the original fee deductor. The system is enabled by setting the related boolean to true by default in `module.go:DefaultGenesis()` which calls `DefaultGenesisState()` on the non AIMD `types/eip1559.go`

The simulation seems to work now, what is currently failing are the e2e tests, I need to investigate.

giunatale:
> The simulation seems to work now, what is currently failing are the e2e tests, I need to investigate.

`test-sim-nondeterminism` is also failing since txs are erroring in the `x/feemarket` ante

https://github.com/atomone-hub/atomone/actions/runs/13905377150/job/38907108528?pr=100

tbruyelle:
> > The simulation seems to work now, what is currently failing are the e2e tests, I need to investigate.
> 
> `test-sim-nondeterminism` is also failing since txs are erroring in the `x/feemarket` ante
> 
> [atomone-hub/atomone/actions/runs/13905377150/job/38907108528?pr=100](https://github.com/atomone-hub/atomone/actions/runs/13905377150/job/38907108528?pr=100)

Oh you're right. Maybe it is related to the e2e error. I noticed that the feemarket module migration was not triggered, resulting in a module without genesis information.

giunatale:
> EDIT: while it is true that the feemarket module migration is not registered, that actually does not affect the current tests because the chain does not need migration when it is created directly from v2. This something else related to the denom resolver I think.

denom resolver? I think I already found what's happening. Selecting any random fee for a tx without matching the minimum gas fee will cause it to fail. Check the logic in `GenAndDeliverTxWithRandFees()`.

giunatale (approved):

PR #101, opened 2025-03-18
chore(x/photon): use `expected_keepers` in `photon/ante`

giunatale:
Also replace usage of deprecated errors APIs with correct module, and fixes some typos.

As per [my comment](https://github.com/atomone-hub/atomone/pull/100#issuecomment-2733078858) in #100, fixing it in `x/photon`.


tbruyelle (changes requested):
The main reason why you would replace an implementation from an other package by a local interface is for testing. Here the tests should be modified in a way that they create a mock from the interface and assert calls on it. Without this change in the tests I don't see the benefit of this change.

Regarding your comment in the feemarket PR, I asked myself twice before replacing the keeper interface with the real keeper type. The reason why I finally chose to use the implementation instead of the interface is because the existing tests did not mock the interface and were actually using [the real implemenation](https://github.com/atomone-hub/atomone/blob/0f0956d62a21b0c1e4993512474bce502328ec01/x/feemarket/ante/suite/suite.go#L39). So basically the interface in the expected_keepers was useless, like the one in this PR.

So for this PR, you should either revert the change regarding the keeper interface or update the tests so they make use of the interface with a mock.

giunatale:
are you implying expected keepers in modules also (generally  defined in types) are also useless and why not include the module keeper in its entirety?

There was a reason this wasn't done imho, and we should stick with that reasoning.

tbruyelle:
OK for isolation, but in any case the tests should take advantage of that interface and mock it.

> are you implying expected keepers in modules (generally defined in types) are also useless and why not include the other modules' keeper in its entirety?

No because these interfaces are all mocked in the module tests.

giunatale:
> OK for isolation, but in any case the tests should take advantage of that interface and mock it.

I understand this part, and yes if we go this route, we should do this too

giunatale:
we need to agree on what direction though, I am voting for updating also tests if anything.

tbruyelle:
> we need to agree on what direction though, I am voting for updating also tests if anything.

let's updating the tests then, you will need to update the makefile as well to create the mocks (see `mocks-gen` target). On my side I will revert the removal of the expected keepers in `ante` and `post` package of `feeemarket`.

PR #103, opened 2025-03-22
chore(x/gov): cap `SensitivityTargetDistance` for dynamic deposit

giunatale:
Also add some comments and logging in case `ApproxRoot` errors.

PR #104, opened 2025-03-22
feat(x/gov): change lazy dynamic deposit computation in favor of endblocker per-tick update

giunatale:
As per the title, it seems unnecessarily complex in hindsight to lazily compute the dynamic deposit.
If we assume per-tick updates aren't too frequent, it's better to do this then only once and limit to reading a single value for queries.
Moreover, since store updates just overwrite an existing value, even a per-block update should not be too expensive in practice.

### DIFF
--- a/x/gov/keeper/min_deposit.go
+++ b/x/gov/keeper/min_deposit.go
@@ -34,8 +34,7 @@ func (keeper Keeper) IncrementActiveProposalsNumber(ctx sdk.Context) {
 	activeProposalsNumber := keeper.GetActiveProposalsNumber(ctx) + 1
 	keeper.SetActiveProposalsNumber(ctx, activeProposalsNumber)
 
-	currMinDeposit := keeper.GetMinDeposit(ctx)
-	keeper.SetLastMinDeposit(ctx, currMinDeposit, ctx.BlockTime())
+	keeper.UpdateMinDeposit(ctx, false)
 }
 
 // DecrementActiveProposalsNumber decrements the number of active proposals by one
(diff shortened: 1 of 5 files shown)


x/gov/keeper/min_deposit.go
```diff
@@ -78,18 +76,33 @@ func (keeper Keeper) GetLastMinDeposit(ctx sdk.Context) (sdk.Coins, time.Time) {
...
+
+// UpdateMinDeposit updates the minimum deposit required for a proposal
+func (keeper Keeper) UpdateMinDeposit(ctx sdk.Context, checkElapsedTime bool) {
 	params := keeper.GetParams(ctx)
-	minDepositFloor := sdk.Coins(params.MinDepositThrottler.FloorValue)
 	tick := params.MinDepositThrottler.UpdatePeriod
+	lastMinDeposit, lastMinDepositTime := keeper.GetLastMinDeposit(ctx)
+	if checkElapsedTime && ctx.BlockTime().Sub(lastMinDepositTime).Nanoseconds() < tick.Nanoseconds() {
+		return
+	}
```
tbruyelle, line 98: There is no additional case in the unit tests for this new condition, is it because it is already covered or is it a miss ?

x/gov/keeper/min_initial_deposit.go
```diff
@@ -80,18 +78,33 @@ func (keeper Keeper) GetLastMinInitialDeposit(ctx sdk.Context) (sdk.Coins, time.
...
+
+// UpdateMinInitialDeposit updates the min initial deposit required for proposal submission
+func (keeper Keeper) UpdateMinInitialDeposit(ctx sdk.Context, checkElapsedTime bool) {
 	params := keeper.GetParams(ctx)
-	minInitialDepositFloor := sdk.Coins(params.MinInitialDepositThrottler.FloorValue)
 	tick := params.MinInitialDepositThrottler.UpdatePeriod
+	lastMinInitialDeposit, lastMinInitialDepositTime := keeper.GetLastMinInitialDeposit(ctx)
+	if checkElapsedTime && ctx.BlockTime().Sub(lastMinInitialDepositTime).Nanoseconds() < tick.Nanoseconds() {
+		return
+	}
```
tbruyelle, line 100: Same question here

tbruyelle:
This PR should be closed in favor of #105 because it already changes the  lazy computation.

giunatale:
I can do that, it's a miss! thanks for spotting it.

both #104 and #103 are included in #105 actually.
I was unsure we were gonna be going with #105 after all but started from #104 as a base.

I will add the coverage there.

PR #107, opened 2025-03-28
test(x/gov): independent cases in KeeperTestSuite

tbruyelle:
State was not reinitialized between test cases, which was confusing for some scenario and forced sometimes to reset things like the deposit throttling.

The fix consists only of calling `suite.reset()` in `KeeperTestSuite.SetupSubTest()`, the other changes removes any correlation previously made by the test cases about this common state.

### DIFF
--- a/x/gov/keeper/grpc_query_test.go
+++ b/x/gov/keeper/grpc_query_test.go
@@ -7,7 +7,6 @@ import (
 
 	"cosmossdk.io/math"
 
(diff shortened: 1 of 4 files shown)


giunatale:
I wanted to do a "lighter" solution tbh with my approach, but this is more rigorous solution. TY

giunatale (approved):

PR #109, opened 2025-03-28
feat(CLI): add `tx simulate` and hide unused flags for `tx broadcast`

### LINKED ISSUE #108: cli: `tx broadcast --dry-run` submits the transaction
While the `tx broadcast` command is supposed to broadcast a signed transaction to the network, when executed with the `--dry-run` flag, it is expected to do it using the simulate mode, but it's actually not. The transaction is submitted.

Investigation in the SDK code revealed that the `--dry-run` flag (which is part of the default tx flags) is ignored. For other commands like `tx gov vote` for instance, it is used to enable the simulation mode, similarily to when `--gas auto` is provided but with the difference that the transacion is not broadcasted after gas estimation.

After discussions with a SDK guy, the behavior is correct; a signed transaction should not be simulated. While the reason was not known, it's maybe for security reason. A simulated signed transaction could be intercepted and executed on the behalf of the original submitter.

The fix is therefore to remove that `--dry-run` flag from the `tx broadcast` command, because by definition only signed transaction can be passed to this command.

In order to simulate a transaction that comes from a JSON file, the [`tx simulate`](https://github.com/cosmos/cosmos-sdk/blob/51086f6ce11d856f7c8a6607efd2541d72a6ab1e/x/auth/client/cli/tx_simulate.go#L18) command could be added. Interestingly the current implementation can take a signed tx but the signature is removed before broadcasting the tx to the network.

Or we could keep `--dry-run` flag but when provided execute a similar behavior than the `tx simulate` command: simulate the tx without the signature.

giunatale:
Closes #108 

Uses the `tx simulate` introduced in Cosmos SDK v0.50.x ([source](https://github.com/cosmos/cosmos-sdk/blob/ede4b3fe238d7839da479c6e2b12e1353f8af073/x/auth/client/cli/tx_simulate.go))

And removes unused and potentially misleading flags from `tx broacast` (simply hides them as they have no effect, except for `FlagOffline` which would cause error, so even better to not show it at all)


giunatale:
Do we want to also reduce the flags shown for `tx simulate`?

tbruyelle:
> Do we want to also reduce the flags shown for `tx simulate`?

`--from` is required so all keyring related flags are required too.

For the others like `account-number`, `sequence`  or `offline`, they are not required but since the command builds a `TxFactory` which always look at these data, it might preferable to keep the flags to avoid unexpected behavior.

tbruyelle (approved):
just a small comment but other than that lgtm

PR #114, opened 2025-04-01
feat: add the x/feemarket module

tbruyelle:
This PR adds the `x/feemarket` module to the main branch, using a fork of https://github.com/skip-mev/feemarket with the following changes:

- compatible with SDK 0.47
- removal of fee tip
- removal of the fee escrowing before distribution

Regarding the integration with AtomOne, the module is configured using the AIMD EIP1559 fee calculation.


proto/atomone/feemarket/v1/tx.proto
```diff
@@ -0,0 +1,33 @@
...
+service Msg {
+  option (cosmos.msg.v1.service) = true;
+
+  // Params defines a method for updating the feemarket module parameters.
+  rpc Params(MsgParams) returns (MsgParamsResponse);
+}
+
+// MsgParams defines the Msg/Params request type. It contains the
+// new parameters for the feemarket module.
+message MsgParams {
```
giuliostramondo, line 22: This message does not seems to be in line with the same type of message in `x/gov` and `x/photon` for the following reasons:
- The name for the param update message used in the other modules is `MsgUpdateParams`.
- The order of the arguments in both the other modules has `authority` in the first position and then `params` in the second position.

tbruyelle: Good point, we should rename to `MsgUpdateParams`, also I noticed the proto declaration lacks the `(amino.name)` option, not sure what's the impact of that but it's probably better to follow the same options than the others, so I would also add
```
  option (amino.name) = "atomone/x/feemarket/v1/MsgUpdateParams";
```
and replicate that change in https://github.com/atomone-hub/atomone/blob/ad9d839df5c3bdd4f182473c1919423dd032be16/x/feemarket/types/codec.go#L14

tbruyelle:
@giunatale  LGTM for your last commit

PR #119, opened 2025-04-07
docs: Updated ADR 002

giuliostramondo:
## Description

Changed ADR 002:

"The only way to `get` PHOTONs" to "the only way to `mint` PHOTONs."


giunatale (changes requested):
I would change it to `mint`

giunatale (approved):

PR #129, opened 2025-04-23
fix: ensures correct denom when minting photon

giuliostramondo:
## fix(x/photon): add denom check on mint transaction

This PR fixes an issue that could arise if a user typed the wrong denom when issuing a photon mint transaction, for example:

`atomoned tx photon mint 1000`**wrongdenom**` -y --home=/path/to/home --fees=330000uatone --keyring-backend=test --output=json  --from=wallet --broadcast-mode=sync --chain-id=chain_name`

Even though the transaction is incorrect, without this fix, the fees would be charged to the user.

### DIFF
--- a/tests/e2e/e2e_exec_test.go
+++ b/tests/e2e/e2e_exec_test.go
@@ -833,7 +833,6 @@ func (s *IntegrationTestSuite) execPhotonMint(
 	valIdx int,
 	from,
 	amt string,
-	gasAuto bool,
 	expectFail bool,
 	opt ...flagOption,
 ) (resp photontypes.MsgMintPhotonResponse) {
@@ -854,9 +853,7 @@ func (s *IntegrationTestSuite) execPhotonMint(
 		"-y",
 	}
 	for flag, value := range opts {
-		if gasAuto || flag != flagGas {
-			atomoneCommand = append(atomoneCommand, fmt.Sprintf("--%s=%v", flag, value))
-		}
+		atomoneCommand = append(atomoneCommand, fmt.Sprintf("--%s=%v", flag, value))
 	}
 
 	if !expectFail {
--- a/tests/e2e/e2e_exec_test.go
+++ b/tests/e2e/e2e_exec_test.go
@@ -762,7 +762,12 @@ func (s *IntegrationTestSuite) expectErrExecValidation(chain *chain, valIdx int,
 		// wait for the tx to be committed on chain
 		s.Require().Eventuallyf(
 			func() bool {
-				gotErr := queryAtomOneTx(endpoint, txResp.TxHash, nil) != nil
+				resp := queryAtomOneTx(endpoint, txResp.TxHash, nil)
+				if isErrNotFound(resp) {
+					// tx not processed yet, continue
+					return !expectErr
+				}
+				gotErr := resp != nil
 				return gotErr == expectErr
 			},
 			time.Minute,
--- a/tests/e2e/e2e_exec_test.go
+++ b/tests/e2e/e2e_exec_test.go
@@ -833,6 +833,8 @@ func (s *IntegrationTestSuite) execPhotonMint(
 	valIdx int,
 	from,
 	amt string,
+	gasAuto bool,
+	expectFail bool,
 	opt ...flagOption,
 ) (resp photontypes.MsgMintPhotonResponse) {
 	opt = append(opt, withKeyValue(flagFrom, from))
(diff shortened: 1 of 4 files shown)


tests/e2e/e2e_exec_test.go
```diff
@@ -833,6 +833,8 @@ func (s *IntegrationTestSuite) execPhotonMint(
 	valIdx int,
 	from,
 	amt string,
+	gasAuto bool,
+	expectFail bool,
 	opt ...flagOption,
```
tbruyelle, line 838: You should be able to control the `--gas` flag using the `opt` vararg. No need for an additionnal arg.

giuliostramondo: the gas fee is set in `execPhotonMint` calling `applyOptions`, it doesn't seem to be in the `flagOption` arg, this is why I added the argument `gasAuto`.

tbruyelle: `opt` is specifically there to override the values initially set in `applyOptions`. I tested with the following change and tests are still green. Again, no need for additionnal `gasAuto` arg.

```diff
--- a/tests/e2e/e2e_photon_test.go
+++ b/tests/e2e/e2e_photon_test.go
@@ -64,13 +64,12 @@ func (s *IntegrationTestSuite) testMintPhoton() {
 		)
 		alice, _ := c.genesisAccounts[1].keyInfo.GetAddress()

-		atoneFees := sdk.NewCoin(uatoneDenom, standardFees.Amount)
 		beforeBalance, err := queryAtomOneAllBalances(chainEndpoint, alice.String())
 		s.Require().NoError(err)

 		// Issue an incorrect transaction with wrong Denom
 		_ = s.execPhotonMint(s.chainA, valIdx, alice.String(), "1000wrongDenom",
-			false, true, withKeyValue(flagFees, atoneFees),
+			false, true, withKeyValue(flagGas, "200000"), // override gas auto
 		)

 		time.Sleep(1 * time.Second)
```

Note that this diff also removes the `withKeyValue(flagFees, atoneFees),` because it passed the standardFees which is the default value, so no need to override it for this test.

x/photon/types/msgs.go
```diff
@@ -47,6 +48,10 @@ func (msg *MsgMintPhoton) ValidateBasic() error {
 	if !msg.Amount.IsPositive() {
 		return errorsmod.Wrapf(sdkerrors.ErrInvalidCoins, "coin to burn must be positive")
 	}
+	// Ensure burned amount denom is bond denom
+	if msg.Amount.Denom != params.BondDenom {
+		return errorsmod.Wrapf(sdkerrors.ErrInvalidCoins, ErrBurnInvalidDenom.Error())
```
tbruyelle, line 53:
```suggestion
		return errorsmod.Wrapf(ErrBurnInvalidDenom, "invalid denom %s", msg.Amount.Denom)
```

tests/e2e/e2e_photon_test.go
```diff
@@ -55,4 +56,38 @@ func (s *IntegrationTestSuite) testMintPhoton() {
 	s.Run("mint photon", subtest(standardFees))
 	atoneFees := sdk.NewCoin(uatoneDenom, standardFees.Amount)
 	s.Run("mint photon with atone fees", subtest(atoneFees))
+	s.Run("mint photon wrong denom", func() {
```
tbruyelle, line 59:
```suggestion
	s.Run("mint photon wrong denom does not deduct fees", func() {
```

tests/e2e/e2e_photon_test.go
```diff
@@ -55,4 +56,38 @@ func (s *IntegrationTestSuite) testMintPhoton() {
...
+		alice, _ := c.genesisAccounts[1].keyInfo.GetAddress()
+
+		beforeBalance, err := queryAtomOneAllBalances(chainEndpoint, alice.String())
+		s.Require().NoError(err)
+
+		// Issue an incorrect transaction with wrong Denom
+		_ = s.execPhotonMint(s.chainA, valIdx, alice.String(), "1000wrongDenom",
+			true, withKeyValue(flagGas, "200000"))
+
+		time.Sleep(1 * time.Second)
```
tbruyelle, line 74: Unecessary, thanks to the `*execValidation` funcs that wait for tx result.

giuliostramondo: I had empirically verified that there is that there is a small (and variable) delay between the moment the execPhotonMint returns and the moment fees are deducted.

Given your comment, I tested this in a more structured manner:
- I removed the denom check in msg.go - so that the test should fail.
- I ran the test with the sleep of 1 second for 10 times, it consistently fails (as expected).
- I ran the test without the sleep of 1 second for 10 times, the result of the test is unreliable, it failed 7 times and passed 3 times.

Therefore, I think we should keep the sleep.

tbruyelle: that makes 2 of my assumptions to be wrong in this PR (this and the non-determistic test...). I need to spend a bit more time on that.

giuliostramondo: I retested this - [here](https://gist.github.com/giuliostramondo/5bef23f1daaeb4e7f3568f97db868f87) is a log - , and obtained the same result (flaky test behavior without the sleep).

tbruyelle: OK, I forgot to comment the `time.Sleep` :roll_eyes:, I can reproduce the flakyness now. What I dont understand is we do the same kind of assertion in the test just above and it works w/o sleep...

tbruyelle: So the reason of the flakyness is because when `expectFails=true`, the error is not necessarily the expected one (wrong denom), but can for instance `404 not found` because the tx is still not yet processed, which leads to premature assertion execution. The fix is to ensure that the error is the one expected.

tbruyelle (approved):

PR #130, opened 2025-04-29
test(e2e): split e2e test setup phase and separated IBC setup

giuliostramondo:
## Description

This PR splits the setup and teardown of the e2e tests into phases. The first setup phase is executed at the beginning of any e2e test and initializes only one chain, without setting up the secondary chain and IBC. The second phase is lazily executed when IBC is needed and performs the IBC setup. 
For most of the tests IBC is not required, and this PR gives substantial speedup on single e2e tests (see below).

The only additional requirement for this PR to work is that any tests that requires IBC to be setup, should start with a call to `s.ensureIBCSetup()`.

Split Setup Phases times:
All e2e tests: 390s
TestPhoton e2e tests: 13s

Current e2e times:
All e2e tests: 469s
TestPhoton e2e tests: 94s

### DIFF
--- a/tests/e2e/e2e_ibc_test.go
+++ b/tests/e2e/e2e_ibc_test.go
@@ -13,7 +13,6 @@ import (
 
 //nolint:unparam
 func (s *IntegrationTestSuite) sendIBC(c *chain, valIdx int, sender, recipient, token, note string) {
-	s.ensureIBCSetup()
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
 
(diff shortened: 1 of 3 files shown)


tests/e2e/e2e_ibc_test.go
```diff
@@ -13,6 +13,7 @@ import (
 
 //nolint:unparam
 func (s *IntegrationTestSuite) sendIBC(c *chain, valIdx int, sender, recipient, token, note string) {
+	s.ensureIBCSetup()
```
tbruyelle, line 16: Did you try to call this only in `(s *IntegrationTestSuite) TestIBC()` method ?

giuliostramondo: Good catch, I just tried your suggestion and it works as expected. I thought that if you tried to execute a specific IBC test (eg. `TestIntegrationTestSuite/TestIBC/send_uatom_to_chainB`) `s.ensureIBCSetup()` might not be called, but I just tested this and it is called.

tbruyelle:
great improvement!

PR #135, opened 2025-05-12
feat(x/gov): dynamic quorum

tbruyelle:
Instead of having a fixed quorum from the parameters, the quorum  is adjusted between 2 bounds, given the proposal vote's participation.

TODO:
- [x] merge main
- [x] migrate `params.LawQuorum` and `params.ConstitutionAmendmentQuorum` to new dynamic model
- [x] fix `types.NewParams` to remove deprecated params
- [x] address `Params` query that returns the dynamic quorum in `TallyParams` but the deprecated quorum value in the `params` struct itself (see also in x/gov/client/cli/query.go)
- [x] Migration to init the values of new parameters (min/max quorum and participation for each kind of kind of proposal)
- [x] use specific participation for constitution amendment and law proposals
- [ ] ~~instead of adding a new Quorum query, consider adding a quorum field to the proposal response (but that could break the compatibility)~~
- [x] add missing CLI for quorum query (if we decide to keep it and not adding new field in proposal response)
- [x] docs: the genesis EMA fields need to be set careful because they will define the initial quorums for the proposals.
- [x] unit tests
- [x] ADR
- [ ] e2e tests
- [ ] Add section in gov module README


proto/atomone/gov/v1/genesis.proto
```diff
@@ -34,10 +35,13 @@ message GenesisState {
   //
   // Since: cosmos-sdk 0.48
   string constitution = 9;
-
+
```
giunatale, line 38:
```suggestion

```

proto/atomone/gov/v1/gov.proto
```diff
@@ -349,4 +349,22 @@ message Params {
...
+  string max_constitution_amendment_quorum = 28 [(cosmos_proto.scalar) = "cosmos.Dec"];
+
+  // Minimum achievable quorum for constitution amendment proposals
+  string min_constitution_amendment_quorum = 29 [(cosmos_proto.scalar) = "cosmos.Dec"];
+
+  // Maximum achievable quorum for law proposals
+  string max_law_quorum = 30 [(cosmos_proto.scalar) = "cosmos.Dec"];
+
+  // Minimum achievable quorum for law proposals
+  string min_law_quorum = 31 [(cosmos_proto.scalar) = "cosmos.Dec"];
```
giunatale, line 369: wouldn't it be better to have a struct with both caps inside for each quorum? (like for dynamic deposit) instead of separate min/max params?

giunatale (changes requested):
I would like to see the EMAs as *separate* per different proposals types, not just the quorum values. So if a proposal is a law or an amendment they contribute to that specific ema otherwise they contribute to the default one

tbruyelle:
> I would like to see the EMAs as _separate_ per different proposals types, not just the quorum values. So if a proposal is a law or an amendment they contribute to that specific ema otherwise they contribute to the default one

Yes, this is part of the TODOs in the PR body.

tbruyelle:
> Moreover, we need a starting value in genesis for the 3 quorums, not just the EMAs (which for a fresh chains are 0). Unless we use the EMAs as starting values, but that's kinda confusing imho. Maybe it could be ok to rely on the EMAs as starting values (since otherwise we would need a separate genesis value and that gets messy probably) but we should make it explicit also in the comments that the EMA should be **non-zero** for a fresh chain and error in case it's 0

Good point. On the other hand, if you start a chain with 0 EMA, you just end up with the minQuorum value for the quorum, that's not catastrophic. But I agree this needs to be well documented. Adding this in the TODO list.

proto/atomone/gov/v1/gov.proto
tbruyelle: Yes, what about embedding the threshold as well ?

tbruyelle:
@giunatale I think we're good with that PR, let's address the e2e tests and the README update later.

proto/atomone/gov/v1/gov.proto
tbruyelle: I created an issue to track the remaining work https://github.com/atomone-hub/atomone/issues/145

giunatale (approved):

PR #137, opened 2025-05-14
chore(x/feemarket): Fixed lt and gt in documentation to reflect the implementation

giuliostramondo:
Description

The documentation AIMD.md is not in line with the [code](https://github.com/atomone-hub/atomone/blob/feat/x/feemarket/x/feemarket/types/state.go#L109):

`	if avg.LTE(params.Gamma) || avg.GTE(math.LegacyOneDec().Sub(params.Gamma)) {`


tbruyelle (approved):
good catch

PR #142, opened 2025-06-06
chore(x/feemarket): update README.md

giunatale:
As per the title, forgot to update the docs to reflect the addition of the new parameter before merging the feemarket PR.
Fixing it here.


tbruyelle (approved):
good catch!

PR #146, opened 2025-06-23
docs: Update gov README with dynamic quorum info

giuliostramondo:
## Description

Update of `x/gov` module README adding dynamic quorum information.

### DIFF
--- a/x/gov/README.md
+++ b/x/gov/README.md
@@ -706,7 +706,6 @@ field, the related proposal will be tallied using specific quorum and threshold 
 instead of the default ones for regular proposals. More specifically, the following parameters
 are added to enable this behavior:
 
-  amendment proposals~
 - `constitution_amendment_threshold` which defines the minimum proportion of Yes votes for a
   Constitution Amendment proposal to pass.
 - `law_threshold` which defines the minimum proportion of Yes votes for a Law proposal to pass.


x/gov/README.md
```diff
@@ -677,12 +706,15 @@ field, the related proposal will be tallied using specific quorum and threshold
 instead of the default ones for regular proposals. More specifically, the following parameters
 are added to enable this behavior:
 
-- `constitution_amendment_quorum` which defines the quorum for constitution amendment proposals
+- ~`constitution_amendment_quorum` which defines the quorum for constitution
+  amendment proposals~
```
tbruyelle, line 710: Let's remove them completely instead no?

tbruyelle (approved):
Some small comments but looks good overall, thank you!

x/gov/README.md
```diff
@@ -320,7 +349,7 @@ which is modifiable by governance. This means that proposals are accepted if:
 
 #### No inheritance
 
-If a delegator does not vote, it won't inherit its validator vote.
+If a delegator does not vote, his validators will not inherit his vote.
```
giunatale, line 352:
```suggestion
If a delegator does not vote, the vote of the delegated validator - if applicable - will not be inherited.
```

giunatale: there's a mistake here, the delegator inherits the validator vote not the other way around

x/gov/README.md
giunatale: agreed

giunatale (changes requested):
added some suggestions.
Overall looks good, but we need to first act on the pending suggestions before I can comfortably approve

x/gov/README.md
tbruyelle: @giuliostramondo There is still one orphan line at line 709.

giuliostramondo: Yup, sorry, good catch

PR #147, opened 2025-06-23
fix(x/gov): add boundary checks for unified diffs

giunatale:
As per PR title, add checks to catch and error on malformed unified diffs (with hunks going outside the actual length of the source document).


tbruyelle (approved):
LGTM

PR #148, opened 2025-06-23
docs(x/gov): dynamic quorum ADR update

giuliostramondo:
## Description

Added details about the implementation of the separate dynamic quorums and regarding the vote participation initialization from genesis.


tbruyelle (changes requested):
Let's wait for the forum post review, we might have other changes to apply to the ADR.

giunatale (approved):

tbruyelle (approved):

Issue #145, opened 2025-06-13
Dynamic quorum improvements


tbruyelle:
- [x] add a common type to contain min and max quorum
- [x] add section in the gov module README
- [x] add e2e tests #154

PR #152, opened 2025-06-24
fix: fixes endpoint address and adds quorums query in gov cli

giuliostramondo:
## Description

Closes: #150 

Renames the rpc endpoints and adds the `quorums` missing query in the gov CLI.

Issue #150, opened 2025-06-24
`x/gov`: add missing quorum query CLI


tbruyelle:
- [ ] Change endpoint path from `/atomone/gov/v1/params/quorum` to `/atomone/gov/v1/quorums`
- [ ] add missing `q gov quorums` CLI command to query that endpoint
- [ ] add `q gov quorums` document in README

PR #154, opened 2025-06-27
test(e2e): dynamic quorum e2e tests

giuliostramondo:
## Description

E2E tests for the dynamic quorum feature

Closes:  #145 
(with PR #146)

Based on PR #152

### DIFF
--- a/tests/e2e/e2e_gov_test.go
+++ b/tests/e2e/e2e_gov_test.go
@@ -325,6 +325,161 @@ func (s *IntegrationTestSuite) testGovConstitutionAmendment() {
 	})
 }
 
+func (s *IntegrationTestSuite) testGovDynamicQuorum() {
+	s.Run("dynamic quorum change", func() {
+
+		chainAAPIEndpoint := fmt.Sprintf("http://%s", s.valResources[s.chainA.id][0].GetHostPort("1317/tcp"))
+		params := s.queryGovParams(chainAAPIEndpoint, "tallying")
+		s.T().Logf("Tally Params: %s", params)
+		quorums := s.queryGovQuorums(chainAAPIEndpoint)
+		// pEma = (Q - Qmin) / (Qmax - Qmin)
+		quorumRange := params.GetParams().QuorumRange
+		quorumMin := sdk.MustNewDecFromStr(quorumRange.Min)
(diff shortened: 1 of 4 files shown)


tbruyelle:
Since #150 has been merged, can you rebase on main so we have a better diff

tests/e2e/e2e_gov_test.go
```diff
@@ -325,6 +325,123 @@ func (s *IntegrationTestSuite) testGovConstitutionAmendment() {
...
+		quorums := s.queryGovQuorums(chainAAPIEndpoint)
+		quorumRange := params.GetParams().QuorumRange
+		quorumMin := sdk.MustNewDecFromStr(quorumRange.Min)
+		quorumMax := sdk.MustNewDecFromStr(quorumRange.Max)
+		currentQuorum := sdk.MustNewDecFromStr(quorums.GetQuorum())
+		quorumPEma := (currentQuorum.Sub(quorumMin)).Quo(quorumMax.Sub(quorumMin))
+		paramsFeemarket := s.queryFeemarketParams(chainAAPIEndpoint)
+		oldAlpha := paramsFeemarket.Params.Alpha
+		paramsFeemarket.Params.Alpha = oldAlpha.Add(sdk.NewDec(1))
+		s.writeFeemarketParamChangeProposal(s.chainA, paramsFeemarket.Params)
```
tbruyelle, line 344: I would rather use a less specific proposal for this test, it's confusing to imply the feemarket in it. We should use a simple text proposal, but there is currently no test for a text proposal submission, right ?

giunatale: agreed, also this will make life easier for the renaming of the module, so we don't have leftover "feemarket" references

giunatale (changes requested):
As also @tbruyelle stated, quote:

> I would rather use a less specific proposal for this test, it's confusing to imply the feemarket in it. We should use a simple text proposal, but there is currently no test for a text proposal submission, right ? Maybe it's time to add one.

This is the only thing I would change here, regarding instead

> Reading this code makes me realize that maybe the participationEMAs should be queryable. Maybe we could return the different participations in the Quorums query response ?

I think this has to be tracked via a separete, dedicated issue and subsequent PR to address it

giunatale (approved):

PR #156, opened 2025-06-30
fix(x/gov): add boundary checks for unified diffs (backport #147)

giunatale:
Backport of #147


giunatale:
this PR would also imply we need to issue a new version I am afraid (and potentially this one is breaking IF the error is triggered)

PR #161, opened 2025-07-07
fix(gov): change `getQuorumAndThreshold` to return highest quorum and threshold

giunatale:
If a proposal contains multiple messages (e.g. law, amendment), always return the highest value for quorum and threshold


tbruyelle (approved):
Looks good, thanks for the fix!

PR #163, opened 2025-07-07
fix(gov): ensure that quorum caps are consistent (max >= min)

giunatale:
Add validation for quorum caps to ensure that max value is greater or equal than min value


tbruyelle (approved):
I added some tests :P, LGTM

Issue #162, opened 2025-07-07
`x/feemarket`: fetch maximum block gas from context instead of consensus param keeper


tbruyelle:
Currently the maximum block gas, used to compute the target block gas, is fetched from the imported `ConsensusParamsKeeper` from the `x/consensus` module.

But this import is actually not required since the max block gas is also available in the passed context via the block gas meter. Furthermore, the maximum block gas inside the block gas meter use an additional logic compared to the `ConsensusParamsKeeper`, defined inside `baseapp.GetMaximumBlockGas()`, which turns a `-1` value into `0`.

For this reason I think it matters to replace the usage of `ConsensParamsKeeper` by `ctx.BlockGasMeter().Limit()`, this removes a dependency to the `x/consensus` module, and it follows a better logic.

tbruyelle:
Unexpected behavior, the value `-1` is actually handled as `math.MaxUint64` because of the type case here

https://github.com/atomone-hub/atomone/blob/833b2502e571cc68119e6ddc6aa5669bcd889f5a/x/feemarket/keeper/feemarket.go#L33

In other words, when `consensusParams.Block.MaxGas=-1`, `maxBlockGas=18446744073709551615`.

PR #178, opened 2025-07-09
fix(x/feemarket): Removes dependency on ConsensusParamKeeper from x/feemarket

giuliostramondo:
## Removes dependency on ConsensusParamKeeper

ConsensusParamKeeper is used by the feemarket module to fetch the max block gas information. 
The use of ConsensusParamKeeper can be replaced with `ctx.ConsensusParams()`, which allows to remove the dependency on `x/consensus`.

Closes: #162

### DIFF
--- a/x/feemarket/post/post_test.go
+++ b/x/feemarket/post/post_test.go
@@ -106,6 +106,7 @@ func TestPostHandle(t *testing.T) {
 					return ctx, nil
 				}
 			)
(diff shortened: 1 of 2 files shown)


giunatale (changes requested):
please start from https://github.com/atomone-hub/atomone/tree/giunatale/rename-feemarket and avoid still using `feemarket` in new/changed code.

tbruyelle:
> please start from [`giunatale`/rename-feemarket](https://github.com/atomone-hub/atomone/tree/giunatale/rename-feemarket) and avoid still using `feemarket` in new/changed code.

Or maybe we can merge the rename first and then rebase?

Issue #164, opened 2025-07-07
`x/feemarket`: handle when block max gas is 0


tbruyelle:
When max block gas is 0, there is no limit in term of consumed gas per block. While it is not the case for Atomone (currently max block gas = `100,000,000`), we might want to handle that case as a precaution.

Especially since this will indirectly disable the `x/feemarket` module by affecting the target block gas which becomes 0 too. Note that this disable is a bit ugly since it comes from a [recovered panic](https://github.com/atomone-hub/atomone/blob/8fee840c2c5be5fc4dc2509d1b690aeaebe2ed6d/x/feemarket/types/state.go#L53-L60). In this particular scenario, we might want to introduce a default value for the target block gas, so the feemarket is still enabled.

PR #180, opened 2025-07-10
fix(x/dynamicfee): handle maxBlockGas in ConsensusParam  set to 0 or -1

giuliostramondo:
## Handles invalid values of maxBlockGas for the dynamicfee module

The new parameter `DefaultMaxBlockGas` is added to the parameters of the dynamicfee module.
If ConsensusParams returns a maxBlockGas that is *0* or *-1*  DefaultMaxBlockGas is used.

Closes: #164


giunatale:
momentarily changing the target back to https://github.com/atomone-hub/atomone/tree/giunatale/rename-feemarket to ease up reviewing. We can change the target back to main once https://github.com/atomone-hub/atomone/pull/170 is merged

x/dynamicfee/post/post_test.go
```diff
@@ -74,6 +74,7 @@ func TestPostHandle(t *testing.T) {
 				expectedState := types.DefaultState()
 				expectedState.Window[0] = gasConsumed
 				m.DynamicfeeKeeper.EXPECT().SetState(m.ctx, expectedState)
+				m.DynamicfeeKeeper.EXPECT().GetMaxGas(m.ctx).Return(uint64(30_000_000), nil)
```
tbruyelle, line 77: Please use `testutil.MaxBlockGas` instead (same for line below).

x/dynamicfee/keeper/keeper_test.go
```diff
@@ -85,3 +86,42 @@ func TestEnabledHeight(t *testing.T) {
 		require.Equal(t, int64(10), got)
 	})
 }
+
+func TestMaxGas(t *testing.T) {
```
tbruyelle, line 90:
```suggestion
func TestGetMaxBlockGas(t *testing.T) {
```

x/dynamicfee/keeper/keeper_test.go
```diff
@@ -85,3 +86,42 @@ func TestEnabledHeight(t *testing.T) {
...
+	k, ctx := testutil.SetupKeeper(t, 0)
```
tbruyelle, line 91: Since `GetMaxBlockGas()` is a pure function (it does not alter state), it's not a problem here, but usually you should create a new keeper for each case to avoid state share between subtests.

tbruyelle (approved):

giunatale (approved):

PR #182, opened 2025-07-13
feat: ADR-004 (Nakamoto Bonus)

### LINKED ISSUE #173: Add parameter to enable and disable the Nakamoto Bonus
We need to be able to quickly shutdown through governance the Nakamoto Bonus feature.

This can be implemented with a parameter `enableNakamotoBonus` in `x/distribution`. If `enableNakamotoBonus` is set to false, the `eta` parameter needs to be set to *0* and the dynamic update of eta needs to be skipped.

Pantani:
implementation of https://github.com/atomone-hub/atomone/blob/main/docs/architecture/adr-004-nakamoto-bonus.md

needs merge first: #181 #141.

Also closes #173.


Pantani:
close because we move this change to atomone cosmos-sdk

https://github.com/atomone-hub/cosmos-sdk/pull/10

Issue #185, opened 2025-07-16
Make participation EMAs queryable


giunatale:
As the title suggests we would like to have the three EMAs queryable in a similar way to the three quorum thresholds, i.e. using a single query endpoint that returns all three EMAs together.

from https://github.com/atomone-hub/atomone/pull/154#discussion_r2194620570

PR #186, opened 2025-07-16
feat(gov): Added query endpoint for participationEMAs

giuliostramondo:
## Added query endpoint for participationEMAs

- Added RPC endpoint `/atomone/gov/v1/participationemas`
- Added cli respective query `atomoned query gov participation`

Closes: #185


giunatale (changes requested):
Can you add a test in `keeper/grpc_query_test.go`?

Overall looks good, just missing a test IMO

giunatale:
also, please do a run of `make proto-pulsar-gen`

PR #190, opened 2025-07-24
fix(gov): fix CLI command `init genesis` param generation for dynamic deposit

giunatale:
This PR fixes the generation of params for the dynamic deposit in `x/gov` when using the CLI command `init genesis` if another denom is passed with the flag `--default-denom`.

The correct denom is now used instead of always picking the default one (`stake`) regardless.


giuliostramondo (approved):
I tested the genesis file generation and the issue with the denom is now solved.

PR #192, opened 2025-07-29
Fixed dynamicfee gas price change e2e test

giuliostramondo:
The gas price change e2e test for the  `dynamicfee` module  has a loop that waits until the value of the learning rate goes below the maximum learning rate.

This loop is incorrect, as the learning rate will increase by waiting (average block utilization goes to 0, which is far from the target block utilization, hence additive increase is used to update the learning rate).

This check was not crucial to the test - that only verifies that there is a gas price increases after a big enough transaction - therefore it can be removed.


tbruyelle:
@giuliostramondo If you keep the learning rate assertion at the end of the test, don't you fear that will re-introduce the test flakiness ?

giuliostramondo:
@tbruyelle absolutely, thank you for spotting this!

PR #201, opened 2025-08-08
test(localnet): add Constitution to localnet genesis

tbruyelle:
This is helpful to test constitution amendments proposals.


tbruyelle:
I prefer to have a real constitution, so testing is closer to production. The file is intentionnaly located in `contrib/localnet` folder, so having it oudated is not a problem IMO. We can update it when necessary if we need to test an other constitution amendment.

I can eventually rename the file to constitution-mock.md or something like that to underline even more the fact that it's a test file.

giunatale:
minor conflict to fix but otherwise LGTM

giunatale (approved):

PR #205, opened 2025-08-11
fix(x/gov): wrong denom for min[Initial]Deposit

tbruyelle:
The migration use the default params to feed the minDeposit and minInitialDeposit, but they are using the sdk.DefaultBondDenom which is `stake`.

The fix simply udpates the denom of each to use `uatone`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 
 ### BUG FIXES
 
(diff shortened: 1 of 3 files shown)


giunatale:
should do this to `main` as well

giunatale (approved):

PR #206, opened 2025-08-14
test(simulation/coredaos): simulation tests for the coredaos module

giuliostramondo:
Simulation tests for the coredaos module

### DIFF
--- a/x/coredaos/simulation/genesis.go
+++ b/x/coredaos/simulation/genesis.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/cosmos/cosmos-sdk/types/module"
 	"github.com/cosmos/cosmos-sdk/types/simulation"
-	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
 
 	"github.com/atomone-hub/atomone/x/coredaos/types"
 )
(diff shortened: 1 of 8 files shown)


x/coredaos/simulation/genesis.go
```diff
@@ -107,7 +113,25 @@ func RandomizedGenState(simState *module.SimulationState) {
...
+		authtypes.DefaultParams(),
+		genesisAccs,
+	)
+
+	// Encode into simState
+	bz, err := simState.Cdc.MarshalJSON(authGenesis)
+	if err != nil {
+		panic(err)
+	}
+	simState.GenState[authtypes.ModuleName] = bz
```
tbruyelle, line 136: I don't think this is correct to alter the `x/auth` genesis like this from an other module, even for simulation. Why are you doing this ?

giuliostramondo: Thank you, I applied the suggestion!

tbruyelle (approved):
LGTM
Regarding the improvememt of ok/failure ratio of the coredaos operation,  I dont think the usage of MsgUpdateParams can help by feeding an empty DAO addresses because the gov proposals are not executed during simulation.

PR #207, opened 2025-08-18
chore: v2.1.0 checksums

tbruyelle:
I have checked that build is reproducible.


giunatale:
should add this to `release\v2.x` also

PR #211, opened 2025-08-21
fix(x/gov): wrong denom for min[Initial]Deposit (backport #205)

tbruyelle:
The migration use the default params to feed the minDeposit and minInitialDeposit, but they are using the sdk.DefaultBondDenom which is `stake`.

The fix simply udpates the denom of each to use `uatone`.


giunatale (approved):

tbruyelle:
I had to remove the CL entry added in an other merged PR

PR #224, opened 2025-09-30
chore(x/photon): clarify impossibility of division by zero in `PhotonConversionRate`

giunatale:
Since I'm sick and tired of people reporting this as an issue because they did not understand the fundamental assumption that the burned token is the denom token and thus some of it need to be staked (i.e. locked) to be able to produce blocks meaning it can't go to zero literally ever, I added a clarification and a panic.
Hopefully this should discourage raising the same (wrong) issue in the future.


giunatale:
failing test is unrelated, so merging anyway.

PR #232, opened 2025-10-15
feat: Gno LC with diff from 07-tendermint

clockworkgr:


clockworkgr:
Added app wiring and re-added / adjusted tests.

Not all tests work/pass as we need to implement https://github.com/cosmos/ibc-go/tree/main/testing for Gno. @tbruyelle ?

Issue #242, opened 2025-11-20
`x/gov`: add migration for v4


tbruyelle:
Since #217, we need to intialize 2 missing parameters: `params.ProposalCancelRatio` and `params.ProposalCancelDest`. 
This has to be done using the global upgrade handler.

PR #248, opened 2025-12-02
refactor!: create x/gov wrapper

julienrbrt:
Create x/gov wrapper based on Atom One SDK

Closes: https://github.com/atomone-hub/atomone/issues/242 and  https://github.com/atomone-hub/atomone/issues/244

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,8 +4,8 @@ go 1.24.5
 
 replace (
 	cosmossdk.io/x/feegrant => cosmossdk.io/x/feegrant v0.1.1
-	cosmossdk.io/x/upgrade => github.com/atomone-hub/cosmos-sdk/x/upgrade v0.1.5-atomone.1.0.20251210200739-4aacf0de5158
-	github.com/cosmos/cosmos-sdk => github.com/atomone-hub/cosmos-sdk v0.50.14-atomone.1.0.20251210200739-4aacf0de5158
+	cosmossdk.io/x/upgrade => github.com/atomone-hub/cosmos-sdk/x/upgrade v0.1.5-atomone.1.0.20251211082612-986998e67618
+	github.com/cosmos/cosmos-sdk => github.com/atomone-hub/cosmos-sdk v0.50.14-atomone.1.0.20251211082612-986998e67618
 	github.com/cosmos/ibc-go/v10 => github.com/cosmos/ibc-go/v10 v10.2.0
 )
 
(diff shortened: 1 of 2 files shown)


tbruyelle:
@julienrbrt If possible I'd like to keep the migrations code, I think it's helpful when you need to write a new one, or when you need to refresh your mind about how one upgrade has been coded.

x/gov/keeper/keeper.go
```diff
@@ -1,268 +1,105 @@
...
-	store := ctx.KVStore(keeper.storeKey)
-	bz := keeper.cdc.MustMarshal(&q)
-	store.Set(types.QuorumCheckQueueKey(proposalID, endTime), bz)
+// GetProposal implements GovKeeper.
+func (keeper *Keeper) GetProposal(ctx sdk.Context, id uint64) (v1.Proposal, bool) {
+	sdkProposal, err := keeper.Keeper.Proposals.Get(ctx, id)
+	if err != nil {
+		return v1.Proposal{}, false
+	}
+	atomoneProposal := v1.ConvertSDKProposalToAtomOne(&sdkProposal)
```
tbruyelle: Oh I didn't realize we need to do that, this gonna bit us each time we'll bring changes to fork proposal type and forgot to apply the same change here.

I've recently discovered an interesting pattern used in `golang.org/x/tools/gopls` that might help us to not forget to apply changes from the fork proposal type.

```
// If the size of token.FileSet changes, this will fail to compile.
const delta = int64(unsafe.Sizeof(v1.Proposal{})) - int64(unsafe.Sizeof(sdkv1.Proposal{}))
var _ [-delta * delta]int
```

So currently it doesn't work because delta = -16, for some reasom the size of `sdkv1.Proposal` is 216 while the size of `v1.Proposal` is 200. 

Aren't they supposed to be the same ?

app/upgrades/v4/upgrades.go
```diff
@@ -23,6 +23,8 @@ func CreateUpgradeHandler(
 			return vm, err
 		}
 
+		// TODO: add gov migrations.
```
tbruyelle: Can you detail the migration needed here plz?

x/gov/keeper/keeper.go
```diff
@@ -1,268 +1,105 @@
...
-// SetLegacyRouter sets the legacy router for governance
-func (keeper *Keeper) SetLegacyRouter(router v1beta1.Router) {
-	// It is vital to seal the governance proposal router here as to not allow
-	// further handlers to be registered after the keeper is created since this
-	// could create invalid or non-deterministic behavior.
-	router.Seal()
-	keeper.legacyRouter = router
+// DecrementActiveProposalsNumber implements GovKeeper.
+func (keeper *Keeper) DecrementActiveProposalsNumber(ctx sdk.Context) {
+	// Active proposal count is tracked implicitly by the x/gov keeper.
```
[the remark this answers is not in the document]
julienrbrt: true!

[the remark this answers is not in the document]
julienrbrt: https://github.com/atomone-hub/atomone/pull/248/commits/f167bdf0be9a4fa2cd970bcdc8921a5d1a2ec337

julienrbrt:
> Here is my first review, I've checked the e2e tests and the failing tests are weird, maybe that hides some problems in the wrapper. I'll have a look as well unless you already found the solution.

It's on my list to check. I was playing with megablock.

app/upgrades/v4/upgrades.go
julienrbrt: https://github.com/atomone-hub/atomone/pull/248/commits/4d840f5c300ed27810309de9966e80a80ded1cdf

PR #250, opened 2025-12-09
fix(photon): use `TruncateInt` to compute `uphotonToMint`

giunatale:
Switch from `RoundInt` to `TruncateInt`.
Any potential issue is very far away in the future, but fixing this now is good for future proofing


julienrbrt (approved):
given it is consensus breaking, does it warrant a changelog?

PR #259, opened 2026-01-07
fix(dynamicfee): add `params.ValidateBasic` to `types\MsgUpdateParams`

giunatale:
perform `msg.Params.ValidateBasic()` when `msg.ValidateBasic()` is invoked for `MsgUpdateParams`


tbruyelle:
@giunatale Can we have a test on this ?

PR #261, opened 2026-01-15
feat(ibc): gno light client module

tbruyelle:
Add a Gno light client as a cosmos module for AtomOne.

Gno uses a slightly different version of tendermint than usual cosmos chains, so this light client module is essentially a modified version of `07-tendermint` light client with updated types.

Revamp of #232 with a smaller fix of proto-gen (basically downgrading the docker image from 17 to 15.2) and w/o the main merge.

Reviewers can only check the difference with the legacy Tendermint light client (`07-tendermint`):

```sh
# assuming this repo and github.com/cosmos/ibc-go are in the same directory:
diff -bur atomone/modules/10-gno/ ibc-go/modules/light-clients/07-tendermint/
```


modules/10-gno/helpers.go
```diff
@@ -0,0 +1,170 @@
...
+		Proposer:   nil,
+	}
+
+	for i, val := range valSet.Validators {
+		key := val.PubKey
+		if key.GetEd25519() == nil {
+			return nil, errorsmod.Wrap(clienttypes.ErrInvalidHeader, "validator pubkey is not ed25519")
+		}
+		gnoValset.Validators[i] = &bfttypes.Validator{
+			Address:          crypto.MustAddressFromString(val.Address),
```
Pantani: This helper is still using panic conversions on relayer. Because these helpers are called during `header/misbehaviour` validation, malformed client messages can crash the node instead of `returningErrInvalidHeader`

modules/10-gno/misbehaviour.go
```diff
@@ -0,0 +1,133 @@
...
+	if misbehaviour.Header2.TrustedHeight.RevisionHeight == 0 {
+		return errorsmod.Wrapf(ErrInvalidHeaderHeight, "misbehaviour Header2 cannot have zero revision height")
+	}
+	if misbehaviour.Header1.TrustedValidators == nil {
+		return errorsmod.Wrap(ErrInvalidValidatorSet, "trusted validator set in Header1 cannot be empty")
+	}
+	if misbehaviour.Header2.TrustedValidators == nil {
+		return errorsmod.Wrap(ErrInvalidValidatorSet, "trusted validator set in Header2 cannot be empty")
+	}
+	if misbehaviour.Header1.SignedHeader.Header.ChainId != misbehaviour.Header2.SignedHeader.Header.ChainId {
```
Pantani, line 80: Can we move it after `misbehaviour.Header1.ValidateBasic()` and `misbehaviour.Header2.ValidateBasic()`, so we can avoid a panic and log the correct error if `SignedHead` is nil?

clockworkgr: Fine by me but just for the record, original tm light client functions the same way:

https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/misbehaviour.go#L47-L85

modules/10-gno/helpers.go
clockworkgr: Which helper? crypto.MustAddressFromString() ?

modules/10-gno/helpers.go
```diff
@@ -0,0 +1,174 @@
...
+				PartsHeader: bfttypes.PartSetHeader{
+					Total: int(sig.BlockId.PartsHeader.Total),
+					Hash:  sig.BlockId.PartsHeader.Hash,
+				},
+			},
+			Type:             bfttypes.SignedMsgType(sig.Type),
+			Height:           sig.Height,
+			Round:            int(sig.Round),
+			Timestamp:        sig.Timestamp,
+			ValidatorAddress: crypto.MustAddressFromString(sig.ValidatorAddress),
```
Pantani: Using `MustAddressFromString` here means a malformed relayed header can panic this path instead of returning `ErrInvalidHeader`. Since this helper already returns error, could we switch to `AddressFromString` and handle the failure error? Same concern for `ProposerAddress` below.

modules/10-gno/helpers.go
```diff
@@ -0,0 +1,182 @@
...
+		NumTxs:     header.NumTxs,
+		TotalTxs:   header.TotalTxs,
+		AppVersion: header.AppVersion,
+		LastBlockID: bfttypes.BlockID{
+			Hash: header.LastBlockId.Hash,
+			PartsHeader: bfttypes.PartSetHeader{
+				Total: int(header.LastBlockId.PartsHeader.Total),
+				Hash:  header.LastBlockId.PartsHeader.Hash,
+			},
+		},
```
Pantani, line 154: and here https://github.com/atomone-hub/atomone/pull/261/changes#r3040457610

modules/10-gno/helpers.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	gnoCommit := bfttypes.Commit{
+		BlockID: bfttypes.BlockID{
+			Hash: commit.BlockId.Hash,
+			PartsHeader: bfttypes.PartSetHeader{
+				Total: int(commit.BlockId.PartsHeader.Total),
+				Hash:  commit.BlockId.PartsHeader.Hash,
+			},
+		},
```
Pantani, line 72: This helper dereferences nested proto fields (`commit.BlockId`, `PartsHeader`) without nil checks. Unlike the upstream `*FromProto` path, a malformed relayed header can panic here before we return `ErrInvalidHeader`. Can we validate the nested messages first and return an error instead?

modules/10-gno/helpers.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+		gnoCommit.Precommits[i] = &bfttypes.CommitSig{
+			ValidatorIndex: int(sig.ValidatorIndex),
+			Signature:      sig.Signature,
+			BlockID: bfttypes.BlockID{
+				Hash: sig.BlockId.Hash,
+				PartsHeader: bfttypes.PartSetHeader{
+					Total: int(sig.BlockId.PartsHeader.Total),
+					Hash:  sig.BlockId.PartsHeader.Hash,
+				},
```
Pantani, line 101: same here https://github.com/atomone-hub/atomone/pull/261/changes#r3040457610

julienrbrt (approved):
tACK

PR #264, opened 2026-01-19
chore(coredaos): update also voting period when proposal is vetoed

giunatale:
Although it has no real effect, users could still vote on vetoed proposal if voting period is not set to be expired when a proposal is vetoed. This PR addresses that

### DIFF
--- a/x/coredaos/keeper/msg_server_test.go
+++ b/x/coredaos/keeper/msg_server_test.go
@@ -686,7 +686,7 @@ func TestMsgServerVetoProposal(t *testing.T) {
 		Id:               1,
 		Status:           govtypesv1.StatusVetoed,
 		FinalTallyResult: &emptyTally,
-		VotingEndTime:    &votingEndTime,
+		VotingEndTime:    &votingEndTime, // will be overwritten in the test
 	}
 	depositPeriodProposal := govtypesv1.Proposal{
 		Title:   "Test Proposal",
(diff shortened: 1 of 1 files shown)


x/coredaos/keeper/msg_server.go
```diff
@@ -379,9 +379,12 @@ func (ms MsgServer) VetoProposal(goCtx context.Context, msg *types.MsgVetoPropos
 	}
 	proposal.Status = govtypesv1.StatusVetoed
 
-	// Since the proposal is veoted, we set the final tally result to an empty tally.
+	// Since the proposal is veoted, we set the final tally result to an empty tally
+	// and the voting period ends immediately
 	emptyTally := govtypesv1.EmptyTallyResult()
 	proposal.FinalTallyResult = &emptyTally
+	blockTime := ctx.BlockTime()
+	proposal.VotingEndTime = &blockTime
```
tbruyelle: The fix is fine but actually the reason why a vote is possible on a vetoed proposal is due to a bug in the `x/gov` wrapper.

When a proposal is vetoed, the `x/coredao` module updates the proposal status and invokes `SetProposal()` on the wrapper, which calls `sdkKeeper.Proposals.Set`. But the wrapper should call `sdkKeeper.SetProposal()`, because it contains some logic regarding an other field `sdkKeeper.VotingPeriodProposals`. Basically if the proposal is no longer in voting period, `SetProposal()` removes it from `sdkKeeper.VotingPeriodProposals`.

This is the reason why a vote is possible on a vetoed proposal, because one of the first check that `keeper.AddVote()` is doing, is to verify if the proposal is present in this `VotingPeriodProposals` field.

I will submit an other PR to fix the wrapper.

PR #263, opened 2026-01-19
chore(upgrades): verify v4 upgrade

julienrbrt:
Closes: https://github.com/atomone-hub/atomone/issues/242

Proceed to do the v4 upgrades verification and implements possible missing migrations.

```sh
ATOMONED_BIN=./build/atomoned-v3 ATOMONED_NEW_BIN=./build/atomoned ./contrib/scripts/upgrade-node.sh
```

---

```
error during handshake: error on replay: collections: encoding error: value decode: proto: wrong wireType = 2 for field QuorumCheckCount
```

PR #267, opened 2026-01-22
fix(deps): downgrade ibc-go to v10.1.1

tbruyelle:
ibc-go v10.2.0 depends on cosmos-sdk v0.53 so we should probably stick
to a version that depends on v0.50.


julienrbrt (commented):
I disagree, given v0.53 is API compatible, with a replace we are gaining the new features and improvements from v10.2.0 (or v10.5.0 in https://github.com/atomone-hub/atomone/pull/263/changes#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L33) while keeping v0.50.

PR #277, opened 2026-02-13
fix(coredaos): fix sim error when oversight DAO cannot veto proposal

giunatale:
PR #275 needs an integration due to the fact that in simulation if the oversight DAO tries to veto a proposal it should not be able to, it errors causing the simulation to fail. This PR addresses this potential allowing the simulation to properly continue.

This PR does not need a CL entry since it's an integration of #275

### DIFF
--- a/x/coredaos/simulation/operations.go
+++ b/x/coredaos/simulation/operations.go
@@ -1,6 +1,7 @@
 package simulation
 
 import (
+	"errors"
 	"math/rand"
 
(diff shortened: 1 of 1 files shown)


julienrbrt:
i just just merged https://github.com/atomone-hub/atomone/pull/263 that contained a similar fix (as it was failing on that branch too). If you prefer this one, you should merge main and remove the other boilerplate.

Pantani (approved):

PR #278, opened 2026-02-15
chore(gov): minor fix to governor converters in wrapper

giunatale:
The `LastStatusChangeTime` field was left out of the conversion. This is a really minor issue due to the fact that the field is for internal use, but for completeness, this PR adds it to the converters.


giunatale:
Due to the nature of the fix, there's really no need for a CL entry.

PR #294, opened 2026-03-12
fix(ante): check nested authz

julienrbrt:
Fix A-7.

### DIFF
--- a/ante/gov_vote_ante.go
+++ b/ante/gov_vote_ante.go
@@ -1,6 +1,7 @@
 package ante
 
 import (
+	"context"
 	"errors"
 
 	errorsmod "cosmossdk.io/errors"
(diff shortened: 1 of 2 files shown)


tbruyelle (approved):

tbruyelle (approved):

tbruyelle:
Looks like the test was useful :P

ante/gov_vote_ante.go
```diff
@@ -61,123 +62,127 @@ func (g GovVoteDecorator) AnteHandle(
...
+func (g GovVoteDecorator) validAuthz(ctx context.Context, execMsg *authz.MsgExec) error {
+	for _, v := range execMsg.Msgs {
+		var innerMsg sdk.Msg
+		if err := g.cdc.UnpackAny(v, &innerMsg); err != nil {
+			return errorsmod.Wrap(atomoneerrors.ErrUnauthorized, "cannot unmarshal authz exec msgs")
+		}
+		// Reject nested MsgExec to prevent bypassing stake checks
+		if msg, ok := innerMsg.(*authz.MsgExec); ok {
+			return g.validAuthz(ctx, msg)
+		}
```
Pantani, line 93: Returning directly means we stop validating the remaining sibling messages in the outer `MsgExec`. Could we recurse, handle any error, and then continue the loop instead? Otherwise, a payload like `nested-valid-exec`, `invalid-vote` can still bypass the stake check

```suggestion
		if msg, ok := innerMsg.(*authz.MsgExec); ok {
		    if err := g.validAuthz(ctx, msg); err != nil {
		        return err
		    }
		    continue
		}
```

julienrbrt: Good catch! Added a test for that too.

Pantani (approved):

giuliostramondo (approved):

PR #300, opened 2026-03-16
feat(v4/upgrade): add migration to fix validators commission to 5%

giunatale:
Closes #288


julienrbrt (approved):

tbruyelle (approved):

app/upgrades/v4/upgrades.go
```diff
@@ -573,3 +578,38 @@ func governorValSharesValueCodec(cdc codec.Codec) collcodec.ValueCodec[sdkgovv1.
...
+	if err := stakingKeeper.SetParams(ctx, params); err != nil {
+		return err
+	}
+
+	// Update all existing validators' commission to 5%
+	validators, err := stakingKeeper.GetAllValidators(ctx)
+	if err != nil {
+		return err
+	}
+	for _, validator := range validators {
```
Pantani, line 600: This migration sets every validator's commission rate to 5%, but leaves MaxRate unchanged. If any existing validator has max_rate < 0.05, the upgrade writes an invalid commission config (rate > max_rate). Can we either fail fast on that case or migrate the validator-level commission bounds together with the rate?

```suggestion
	for _, validator := range validators {
		// leaving MaxRate and MaxChangeRate unchanged as validator-defined values.
		validator.Commission.Rate = fivePercent
```

giunatale: shouldn't the commission change fail? I don't understand

giunatale: oh ok I see now, I can upgrade the maxRate if is less than 5%, but only in that case

tbruyelle: Good catch Danilo, this makes me realize that if an `MsgUpdateParams` passes and there is changes in the commission rate range, this should affect the validators as well, just like in this migration.

WDYT? (not asking to do this change here, but if you're agree I'll create a separate issue)

giunatale: makes sense to me yes

tbruyelle: OK created the issue there https://github.com/atomone-hub/cosmos-sdk/issues/75

Pantani (approved):

PR #302, opened 2026-03-17
test(e2e): fix ts-relayer command

tbruyelle:
`add-mnemonic` has changed and now requires a flag.

Also fixed a wrong parameter order in require.Equal.


Pantani (approved):

giuliostramondo (approved):
LGTM

PR #307, opened 2026-03-19
fix(x/coredaos): missing `MsgAnnotateProposal` annotation length validation

tbruyelle:
Fix A-39

### DIFF
--- a/x/coredaos/types/msgs.go
+++ b/x/coredaos/types/msgs.go
@@ -7,6 +7,10 @@ import (
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
+const (
+	MaxAnnotationLength = 1000
+)
+
 var _, _, _, _, _ sdk.Msg = &MsgAnnotateProposal{}, &MsgEndorseProposal{}, &MsgExtendVotingPeriod{}, &MsgVetoProposal{}, &MsgUpdateParams{}
 
 // NewMsgAnnotateProposal creates a new MsgAnnotateProposal instance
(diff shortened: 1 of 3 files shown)


x/coredaos/types/msgs.go
```diff
@@ -7,6 +7,10 @@ import (
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
+const (
+	MaxAnnotationLength = 1000
```
giunatale: Is it enough?
I know it's simpler this way, but also maybe a constant is not ideal and should be a param instead. Just thinking out loud here, thoughts?

tbruyelle: Most length limits are constants, so I did the same here. Do you think it is different for a proposal annotation ?

Pantani (approved):

x/coredaos/types/msgs.go
tbruyelle: @giunatale ping

giunatale: oh sorry I lost this comment. I am not sure anyway, an annotation might be a lengthy review for example, but I guess in that case it would be better to have a link to an external resource after all.
A finite constant might be enough, but maybe the value should be tweaked to allow enough text to be included. 1000 is ok perhaps, but maybe something like 5000? maybe it's too much?

tbh I am not sure, WDYT?

tbruyelle: OK for 5000, I agree that 1k is probably too limited. [4dd6c77](https://github.com/atomone-hub/atomone/pull/307/commits/4dd6c7718da80c1250005fb6b214ff39ad11f92a)

julienrbrt (approved):

PR #308, opened 2026-03-19
fix(x/gov): missing EMA fields validation in genesis

tbruyelle:
Fix A-41

### DIFF
--- a/x/gov/types/v1/genesis.go
+++ b/x/gov/types/v1/genesis.go
@@ -6,6 +6,8 @@ import (
 
 	"golang.org/x/sync/errgroup"
 
+	"cosmossdk.io/math"
+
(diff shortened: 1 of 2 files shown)


giunatale (approved):

Pantani (approved):
One question. These checks are added in the AtomOne wrapper types, but the `app/app.go` still validates through `gov.NewAppModuleBasic(...)` from `github.com/cosmos/cosmos-sdk/x/gov`. Do we need to change it?

tbruyelle:
> One question. These checks are added in the AtomOne wrapper types, but the `app/app.go` still validates through `gov.NewAppModuleBasic(...)` from `github.com/cosmos/cosmos-sdk/x/gov`. Do we need to change it?

Yes SDK fork is changed as well, see https://github.com/atomone-hub/cosmos-sdk/pull/77

PR #309, opened 2026-03-19
fix(x/gov): oversight DAO and steering DAO cannot be the same

giunatale:
Fix A-42

### DIFF
--- a/x/coredaos/types/params.go
+++ b/x/coredaos/types/params.go
@@ -56,6 +56,11 @@ func (p Params) ValidateBasic() error {
 		}
 	}
 
+	// Oversight DAO cannot be the same as the Steering DAO
+	if p.SteeringDaoAddress != "" && p.OversightDaoAddress != "" && p.SteeringDaoAddress == p.OversightDaoAddress {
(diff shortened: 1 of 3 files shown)


tbruyelle (approved):

x/coredaos/types/params.go
```diff
@@ -56,6 +56,11 @@ func (p Params) ValidateBasic() error {
 		}
 	}
 
+	// Oversight DAO cannot be the same as the Steering DAO
+	if p.SteeringDaoAddress != "" && p.OversightDaoAddress != "" && p.SteeringDaoAddress == p.OversightDaoAddress {
```
Pantani: This only compares the two raw strings. `sdk.AccAddressFromBech32` normalizes uppercase Bech32 addresses, so the same underlying account can still be used for both DAOs if one address is uppercase and the other is lowercase. We should compare the decoded `sdk.AccAddresss`, or normalize both strings first

tbruyelle: This yet an other good point Danilo, this needs to be fixed. cc @giunatale

tbruyelle (approved):

Pantani (approved):

PR #315, opened 2026-03-30
chore(v4/upgrade): remove references to cancel proposal

giunatale:
finalize https://github.com/atomone-hub/cosmos-sdk/pull/85 by removing references to the params related to cancel proposal in the v4 migration


tbruyelle (approved):

julienrbrt (approved):
We should make sure to re-tag the SDK and bump it here prior to merge this.

Pantani (approved):

PR #316, opened 2026-03-31
feat(ante): prevent oversight DAO change to be bundled in proposals

### LINKED ISSUE #313: OversightDAO address change makes a proposal impossible to veto.
The anti-veto check scans proposal messages for MsgUpdateParams that modify `OversightDaoAddress`. If it finds one — regardless of what the new address is — it blocks the veto. So you could set the Oversight DAO address to literally any other valid address (doesn't need to be the gov module), and the proposal becomes unvetable. The anti-veto protection was designed to prevent the DAO from blocking its own replacement, but it's too broad — it protects the entire proposal, not just the address-change message.

We should make the anti-veto check more targeted — only block veto of proposals that exclusively change the OversightDAO address.

giunatale:
closes #313


tbruyelle (approved):

giuliostramondo (approved):

ante/gov_submit_proposal_ante.go
```diff
@@ -0,0 +1,130 @@
...
+			proposalMsgs = msg.GetMessages()
+		default:
+			continue
+		}
+
+		if err := g.validateProposalMessages(ctx, proposalMsgs); err != nil {
+			return err
+		}
+	}
+	return nil
```
Pantani: This only checks top-level `MsgSubmitProposal` messages. Since authz is enabled, and `GovVoteDecorator` already unwraps `authz.MsgExec`, a grantee can submit the same bundled proposal through `MsgExec` and bypass this. 

```suggestion
	for _, m := range msgs {
		switch msg := m.(type) {
		case *govv1.MsgSubmitProposal:
			if err := g.validateProposalMessages(ctx, msg.GetMessages()); err != nil {
				return err
			}
		case *sdkgovv1.MsgSubmitProposal:
			if err := g.validateProposalMessages(ctx, msg.GetMessages()); err != nil {
				return err
			}
		case *authz.MsgExec:
			for _, anyMsg := range msg.Msgs {
				var innerMsg sdk.Msg
				if err := g.cdc.UnpackAny(anyMsg, &innerMsg); err != nil {
					return errorsmod.Wrap(atomoneerrors.ErrUnauthorized, "cannot unmarshal authz exec msgs")
				}
				if err := g.ValidateSubmitProposalMsgs(ctx, []sdk.Msg{innerMsg}); err != nil {
					return err
				}
			}
		}
	}
	return nil
```

giunatale: yes thanks was discussing with Giulio, we are aware

ante/gov_submit_proposal_ante.go
```diff
@@ -0,0 +1,170 @@
...
+				var inner sdk.Msg
+				if err := g.cdc.UnpackAny(anyInner, &inner); err != nil {
+					continue
+				}
+				if err := validateMsg(inner); err != nil {
+					return err
+				}
+			}
+			continue
+		}
```
tbruyelle: Is it possible to mutualize the code with `ante/gov_vote_ante.go` ? Looks like we do the same thing with authz messages.

giunatale: can you explain what you mean? not sure I understand "mutualize"

tbruyelle: Both files unwraps messages from authz.MsgExec, and the code is pretty similar.

Sorry for this frenglish "Mutualize" word, but what I mean is changing both files so that some code is shared between them.

giunatale: done

tbruyelle (approved):

PR #317, opened 2026-03-31
fix(x/coredaos): add tests to validate `UpdateParams` execution

Pantani:


x/coredaos/keeper/msg_server.go
```diff
@@ -31,6 +31,10 @@ func NewMsgServer(k *Keeper) types.MsgServer {
 func (ms MsgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParams) (*types.MsgUpdateParamsResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	if err := msg.ValidateBasic(); err != nil {
```
julienrbrt: The code uses ValidateBasic, so revalidation is not necessary: https://github.com/atomone-hub/atomone/blob/main/x/coredaos/types/msgs.go#L139-L146

julienrbrt (changes requested):

x/coredaos/keeper/msg_server.go
```diff
@@ -31,6 +31,10 @@ func NewMsgServer(k *Keeper) types.MsgServer {
...
+		return nil, err
+	}
+
```
Pantani:
```suggestion
```

julienrbrt:
This test is not necessary because this is tested is msgs_test.

PR #320, opened 2026-04-20
fix(coredaos): byte-level check for addresses

giunatale:
Replaces string-level comparisons (which are case sensitive) with the more correct byte-level comparison by converting to `AccAddress`.

Moreover, just for extra caution, normalize the addresses stored in the params struct when `MsgUpdateParams` is invoked.

For the check in `VetoProposal`: empty string (or in practice even a malformed address) would cause an error during parsing, but  the validation done beforehand should prevent an invalid address so the only realistic case in which an error is returned is the empty string. In any case, an address returning a parsing error would be by definition different from the existing one (which must be valid and not empty), and that's why the code is written like that for compactness


tbruyelle (approved):
+1

PR #321, opened 2026-04-21
fix(coredaos): prevent oversight DAO to veto a `authz.MsgExec`-wrapped proposal trying to replace it

giunatale:
This PR addresses the scenario where a proposal replacing the oversight DAO is - for some reason - wrapped in an `authz.MsgExec`. In such case it would be possible for the DAO to veto a proposal replacing it, bypassing the veto prevention mechanism.

However:
- Such proposal would be practically malformed: would require grantee==granter==authority (and in itself this is kinda weird to see, and definitely would not make sense for an honest proposer even though the proposal would be formally valid).
- There's no practical attack scenario that could make use of this inconsistency: no honest proposer wanting to replace the oversight DAO would wrap the param update in a `authz.MsgExec`; conversely, someone malicious (and it could only be the oversight DAO itself or someone affiliated) that submits such a proposal would only be able to somewhat annoy the chain governance, but the effect would be marginal (just to create a proposal, then wait some time during which others might vote on it, before vetoing it), resulting likely in the future replacement of the oversight DAO for having acted in bad faith.

The PR still fixes the code for completeness (to mirror what's done in the ante) but there's no inherent risk nor practical attack that could make use of the issue being fixed here.

### DIFF
--- a/ante/gov_ante_utils.go
+++ b/ante/gov_ante_utils.go
@@ -35,4 +35,3 @@ func iterateMsg(cdc codec.BinaryCodec, msgs []sdk.Msg, fn func(sdk.Msg) error) e
 	}
 	return nil
 }
-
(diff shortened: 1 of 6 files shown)


tbruyelle (approved):

julienrbrt (approved):
utACK

PR #322, opened 2026-04-21
chore(x/dynamicfee): removed the dynamicfee module from atomone

giuliostramondo:
This PR removes the dynamicfee module from atomone and replaces it with the respective cosmos-sdk module (see https://github.com/atomone-hub/cosmos-sdk/pull/48).


julienrbrt (approved):
utACK! 

We should make sure that script still runs fine: https://github.com/atomone-hub/atomone/blob/main/contrib/scripts/upgrade-node.sh prior to merging this.

```
ATOMONED_BIN=./build/atomoned-v3 ATOMONED_NEW_BIN=./build/atomoned ./contrib/scripts/upgrade-node.sh
```

.build.sh
tbruyelle, line 1: I've never noticed this file before, it has never been changed since 5 years, we can probably remove it.

BTW what are these file mode changes everywhere?

giuliostramondo: Addressed both issues in https://github.com/atomone-hub/atomone/pull/322/commits/f21f1c39e7c9252d957d9e67240523fce12e2899 .

tbruyelle:
> utACK!
> 
> We should make sure that script still runs fine: [`main`/contrib/scripts/upgrade-node.sh](https://github.com/atomone-hub/atomone/blob/main/contrib/scripts/upgrade-node.sh) prior to merging this.
> 
> ```
> ATOMONED_BIN=./build/atomoned-v3 ATOMONED_NEW_BIN=./build/atomoned ./contrib/scripts/upgrade-node.sh
> ```

Tested on my machine, the upgrade went well and I can query the dynamicfee module with the new binary (and so the new updated proto path)

tbruyelle (approved):

giunatale:
we'd probably want a changelog entry for this one

giunatale (approved):

PR #328, opened 2026-04-24
fix(v4/upgrade): add participation EMAs migrations

### LINKED ISSUE #325: Gov v4 migration corrupts the store
After v4 upgrade it's no longer possible to query the gov params:

```
$ ./build/atomoned --home ~/.atomone-localnet/  q gov params
Error: rpc error: code = Unknown desc = collections: encoding error:
    value decode: math/big: cannot unmarshal "0.581818181818181818" into a *big.Int: panic
```

### How to reproduce:
- build a v3.3.0 binary
- build a binary from main
- run the `upgrade-node.sh` script, which runs a v3 chain, submit an upgrade proposal, waits for it, switch the binary and run the chain again with v4.
```
ATOMONED_BIN=~/.local/bin/atomoned-v3.3.0 ATOMONED_NEW_BIN=./build/atomoned ./contrib/scripts/upgrade-node.sh
```
- once the upgrade is done, query the gov params using the v4 binary

giunatale:
Closes #325


julienrbrt (approved):
utACK nice find.

tbruyelle (approved):
I confirm that #325 is fixed, well done!

PR #326, opened 2026-04-24
chore(v4/upgrade): apply Option C dynamicfee params in v4 migration

tbruyelle:
Closes #285

PR #330, opened 2026-04-30
fix(v4/upgrade): re-initialize state after param changes in `initDynamicfeeParams`

giunatale:
In #326 the code missed adding an important piece that would create a panic if state is exported and then imported again, and in any case an inconsistency that may also cause unexpected behavior:

The param changes where only set in the store with `SetParams`, but where not actualized in the same way done by `UpdateParams`. 

This PR adds the missing `NewState` plus `SetState` required to complete #326


tbruyelle (approved):
good catch!

PR #333, opened 2026-05-13
feat(v4/upgrade): add `10-gno` to IBC allowed clients

tbruyelle:
Append the `10-gno` light client type to the 02-client `AllowedClients` param during the v4 upgrade so gno clients can be created on-chain.


julienrbrt (approved):
utACK

giunatale (approved):

PR #348, opened 2026-06-23
chore: bump cosmos-sdk to v0.500.1 + CHANGELOG

tbruyelle:
Compat with consensus key rotation:
- Satisfy StakingHook interface as a no-op in x/coredaos
- Add migration for the new staking params rotationFee.
- Add distribution keeper to staking keeper

### DIFF
--- a/app/upgrades/v4/upgrades.go
+++ b/app/upgrades/v4/upgrades.go
@@ -625,8 +625,8 @@ func MigrateStakingParams(ctx context.Context, stakingKeeper *stakingkeeper.Keep
 	params.MaxCommissionRate = fivePercent
 	params.MinCommissionRate = fivePercent
 
-	// Initialize the consensus pubkey rotation fee to 1 ATONE, denominated in the chain bond denom.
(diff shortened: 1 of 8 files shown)


giunatale (approved):

app/upgrades/v4/upgrades.go
```diff
@@ -621,6 +624,10 @@ func migrateValidatorsCommission(ctx context.Context, stakingKeeper *stakingkeep
 
 	params.MaxCommissionRate = fivePercent
 	params.MinCommissionRate = fivePercent
+
+	// Initialize the consensus pubkey rotation fee to 1 ATONE, denominated in the chain bond denom.
```
julienrbrt: any reason to not put it way more expensive? this operations imho should not happen so often, and we try should to avoid a validator rotating key each block

tbruyelle: Thanks for the remind, I raised the value to 100 ATONEs 8bade79e5a52e0a91f2fd49cb29f049d98a6ea27

julienrbrt (approved):

PR #351, opened 2026-06-23
feat: add testnet-specific upgrade handler

### LINKED ISSUE #349: v4/testnet: need a specific release
The current v4 version deployed on the testnet lacks a migration added with the consensus key rotation (see [0]) before the mainnet release.

This means the testnet will require a specific upgrade with only that missing migration, with a specific binary. Or else the parameter won't be initialized properly.

[0]: https://github.com/atomone-hub/atomone/pull/348/changes#diff-e81f74120ece65660c4fc4c115785c9ecee095033395edd659155fafce912a88R629

giunatale:
Closes #349 

Add a `4_1` upgrade handler to initialize the `KeyRotationFee` param in `x/staking`.

Once this is merged, the target `release/v4.x-testnet` branch should be used to create the testnet-specific release

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -59,7 +59,6 @@ import (
 	atomoneante "github.com/atomone-hub/atomone/ante"
 	"github.com/atomone-hub/atomone/app/keepers"
 	"github.com/atomone-hub/atomone/app/upgrades"
-	v4 "github.com/atomone-hub/atomone/app/upgrades/v4"
 	v41 "github.com/atomone-hub/atomone/app/upgrades/v4_1"
 	"github.com/atomone-hub/atomone/client/docs"
 	atomonepost "github.com/atomone-hub/atomone/post"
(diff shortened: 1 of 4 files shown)


app/app.go
```diff
@@ -68,7 +69,7 @@ var (
 	// DefaultNodeHome default home directories for the application daemon
 	DefaultNodeHome string
 
-	Upgrades = []upgrades.Upgrade{v4.Upgrade}
+	Upgrades = []upgrades.Upgrade{v4.Upgrade, v41.Upgrade}
```
tbruyelle: Not sure if this has any impact but why did you keep the `v4.Upgrade` ?

tbruyelle (approved):

app/app.go
giunatale: it was a mistake, I removed it

PR #352, opened 2026-06-25
fix(photon): prevent bundling the wildcard symbol in `TxFeeExceptions` with other exceptions

giunatale:
Having the wildcard symbol `*` bundled with other messages is semantically wrong and should be prevented. The wildcard should exist by itself, and up until this PR an ill-formatted assignment for `TxFeeExceptions` of the kind `[MsgFoo, *]` would silently be accepted and result in the wildcard being ignored, since the code expects it to be the first (and essentially only) entry in the list.

Note: this is a nitpick fix, **it is not required for v4** and can be treated independently and scheduled to be deployed in a future update. There's no urgency nor any cause of concern. This PR is just for code correctness, and impact is marginal at best (also because we don't expect the `TxFeeExceptions` param to be updated any time soon)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # CHANGELOG
 
+## Unreleased
+
+### API BREAKING
+
+### BUG FIXES
+
+- Prevent bundling the wildcard symbol in `TxFeeExceptions` with other exceptions in `x/photon` [#352](https://github.com/atomone-hub/atomone/pull/352)
+
+### DEPENDENCIES
+
+### FEATURES
+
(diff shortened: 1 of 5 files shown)


giunatale:
BTW, if we want a CL entry for this, I'll need to go ahead and add a new section above v4. LMK

tbruyelle (approved):
Yes let's add a new section for the CL.

giunatale:
Ok done @tbruyelle if it looks good I'll merge. It should be ok, but asking for confirmation just in case

tbruyelle:
yeah it's good go ahead

PR #353, opened 2026-06-26
refactor(coredaos): migrate away from atomone's `x/gov` wrapper

giunatale:
This PR removes the dependency on atomone `x/gov` wrapper from `x/coredaos` in favor of using the atomone-sdk `x/gov`.

This work alongside https://github.com/atomone-hub/atomone-sdk/pull/106 is a prerequisite for a future-coming PR that does some further cleaning/refactoring.


tbruyelle (approved):
Having to deal direclty with the gov keeper fields like `keeper.Proposals` is a bit annoying as it forbids the usage of a local interface. But that's out of the scope of this PR.
LGTM

giunatale:
> Having to deal direclty with the gov keeper fields like `keeper.Proposals` is a bit annoying as it forbids the usage of a local interface. But that's out of the scope of this PR.
> LGTM

nothing to do here unfortunately because of collections, using the keeper directly is unavoidable. The methods the previous interface was using are simply gone

PR #356, opened 2026-07-22
fix: register legacy IBC gov proposal content types in atomone gov fork

tbruyelle:
The v4 mainnet upgrade failed during the gov proposal migration with:

  failed to migrate gov proposals: failed to get proposal value: no
  concrete type registered for type URL /ibc.core.client.v1.ClientUpdateProposal
  against interface *v1beta1.Content

The AtomOne gov module fork defines its own Content interface (atomone.gov.v1beta1.Content), and MsgExecLegacyContent unpacks its content Any against it. ibc-go registers ClientUpdateProposal and UpgradeProposal against the SDK's gov Content interface, but not against ours, so a historical mainnet proposal (#20) wrapping a ClientUpdateProposal failed to unpack during migration.

Register both legacy IBC 02-client proposal content types against the AtomOne Content interface, matching the existing manual registration of the params and upgrade proposal content types.

An audit of all 20 atomone-1 mainnet proposals confirms ClientUpdateProposal is the only missing registration; UpgradeProposal is registered defensively as it travels with it in ibc-go.


giunatale (approved):

julienrbrt (approved):
utACK

julienrbrt:
We should backport this to main

PR #364, opened 2026-08-24
fix(dynamicfee): consume failed-tx gas in dynamic fee pricing

giunatale:
See https://github.com/atomone-hub/atomone-sdk/pull/110

Now that gas accounting is authoritative in the SDK's `EndBlock`, the redundant `x/dynamicfee` post handler is removed from the app wiring.


giunatale:
This PR requires an update of the go.mod once https://github.com/atomone-hub/atomone-sdk/pull/110 is merged (and new tag/version issued) on the atomone-sdk.

It will also likely require backporting on `release/v4.x` and the issuance of a new version

PR #365, opened 2026-08-24
fix(coredaos): defer MsgVetoProposal deposit/vote cleanup to EndBlocker

giunatale:
The `x/coredaos` `MsgVetoProposal` refunded/burned every deposit and deleted every vote of the target proposal inline, inside the message handler. That handler runs under the metered transaction gas meter, which is itself bounded by `block.max_gas`. Both loops are unbounded and scale linearly with the number of depositors and votes, quantities anyone can inflate during the voting period.

Once the cumulative cost exceeds `block.max_gas` (100,000,000 on mainnet), no veto transaction for that proposal can be included in any block, and the Oversight DAO's constitutional veto becomes permanently unavailable for it.

## Fix

`x/gov`'s EndBlocker (which already performs exactly these deletions for rejected proposals under the infinite block gas meter) lives in the AtomOne-SDK, so deferring there would be a two-repo change. Moreover, while the fix affect proposals, it's related to behavior implemented in the `x/coredaos` module. For these reasons the fix stays entirely within `x/coredaos`:

- `VetoProposal` is now O(1): it marks the proposal `Vetoed`, resets the tally,
  ends voting immediately, removes the proposal from the gov active queue (so
  gov never tallies or executes it), and enqueues its id for cleanup.
- A new module `EndBlocker` drains that queue, performing the (potentially
  unbounded) deposit refund/burn and vote deletion. `EndBlock` runs under the
  infinite block gas meter, exactly like gov's own cleanup of rejected
  proposals, so it can never be priced out of a block.


x/coredaos/keeper/abci.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+		// Delete all votes for the proposal. Votes are stored as
+		// collections.Map[collections.Pair[uint64, sdk.AccAddress], v1.Vote].
+		if err := k.govKeeper.Votes.Clear(ctx, collections.NewPrefixedPairRange[uint64, sdk.AccAddress](proposalID)); err != nil {
+			return errors.Wrapf(err, "error deleting votes for vetoed proposal %d", proposalID)
+		}
+		if err := k.VetoCleanupQueue.Remove(ctx, proposalID); err != nil {
+			return errors.Wrapf(err, "error removing vetoed proposal %d from cleanup queue", proposalID)
+		}
+	}
```
tbruyelle, line 63: A cleanup error can now halt the chain. We should use a cached context and only log in case of error.

x/coredaos/keeper/abci.go
```diff
@@ -0,0 +1,60 @@
...
+// VetoProposal only marks a proposal as vetoed and enqueues its id; the
+// potentially unbounded work — refunding or burning every deposit and deleting
+// every vote — is carried out here. EndBlock runs under the infinite block gas
+// meter, so this cleanup cannot be priced out of a block the way it could if it
+// ran inside the metered veto transaction. This mirrors how x/gov cleans up
+// rejected proposals from its own EndBlocker.
+//
+// The queue is drained fully every block, so it is always empty at a block
+// boundary and needs no genesis import/export.
+func (k Keeper) EndBlocker(ctx context.Context) error {
```
tbruyelle: Because gov `EndBlocker` is executed before coredaos `EndBlocker`, there is a slight chance that a vetoed proposal is resurrected: `QuorumCheckQueue` may insert it again in `ActiveProposalQueue`.

Moving the core `EndBlocker` before the gov `EndBlocker` should fix the problem.

x/coredaos/keeper/abci.go
```diff
@@ -0,0 +1,60 @@
...
+	// Collect the ids first: the cleanup callbacks below must not mutate the
+	// VetoCleanupQueue while it is being walked.
```
tbruyelle: I don't think this is true. Look at `DeleteAndBurnDeposits` or `RefundAndDeleteDeposits`, they both remove items inside the walk func.

tbruyelle (changes requested):

x/coredaos/keeper/abci.go
giunatale: Good catch, thank you — and this was a regression from deferring the cleanup. With the old inline code the votes were deleted at tx time, before any EndBlocker ran; deferring them meant gov's `QuorumCheckQueue` walk could still observe them, pass quorum, and re-insert the proposal into `ActiveProposalsQueue`. Fixed in f938ac27 by ordering the coredaos EndBlocker before gov's in `orderEndBlockers`, so a vetoed proposal's votes are gone before the quorum check runs. Added `TestCoreDAOsEndBlockerRunsBeforeGov` to guard the ordering and documented the requirement on the EndBlocker.

x/coredaos/keeper/abci.go
giunatale: You're right, thanks. `RefundAndDeleteDeposits`/`DeleteAndBurnDeposits` both remove inside their walk, and gov's own EndBlocker even removes from and re-adds to `QuorumCheckQueue` mid-iteration — so the collect-first indirection and its comment were unnecessary. Removed in e3194019: the proposal is now removed from the queue inline within the walk.

x/coredaos/keeper/abci.go
giunatale: Agreed. Fixed in 50c6e14d: each proposal's cleanup now runs in its own `ctx.CacheContext()` — committed on success, and on error we log and leave the entry queued to retry next block, so a cleanup failure can no longer abort the block.

tbruyelle (approved):

julienrbrt (approved):
utACK

PR #366, opened 2026-09-10
fix(x/gov): decode pre-v4 gov params in historical queries

clockworkgr:
## Description

Closes: none. Found while indexing atomone-1 from genesis against an archive node.

The v4 upgrade moved gov params to the `cosmos.gov.v1` layout of the SDK fork, which renumbers every field from 15 upwards by one. A historical query runs the current binary against pre-migration state, so params written by v1, v2 and v3 are read with the wrong type:

| State era | Heights | What the current type reads |
|---|---|---|
| v1, v2 | below 5,902,000 | decodes without error, one field off: `min_deposit_ratio` dropped, every later field shifted, quorum ranges nil |
| v3 | 5,902,000 to 9,549,999 | fails: `wrong wireType = 2 for field QuorumCheckCount` (legacy field 23 is a message, the new field 23 a varint) |

The nil quorum ranges then panic in `GetQuorum`. On atomone-1 every `TallyResult` and `Params` query at a height inside a proposal's voting period before the v4 upgrade returns ABCI code 111222 or 18. Reproduction with nothing but curl:

```
https://atomone-rpc-history.infra.allinbits.vip/abci_query?path=%22/atomone.gov.v1.Query/TallyResult%22&data=0x0801&height=749400
```

This change gives the gov keeper's `Params` collection a layout-aware codec (`x/gov/keeper/params_codec.go`), installed in `app/keepers/keepers.go` right after the keeper is built:

- the pre-v4 layout is recognised by its field 15 marker (`min_deposit_ratio`), which the current layout never writes;
- legacy bytes are decoded with the retained `atomone.gov.v1` type and converted with the existing `ConvertAtomOneParamsToSDK`, the same path the v4 migration used;
- state that predates the dynamic quorum gets quorum ranges pinned to its static quorum, so the dynamic-quorum code returns exactly the quorum the chain applied then;
- encoding and current-layout decoding are unchanged.

The tests decode the actual params bytes read from atomone-1 state at heights 749,400, 6,085,557 and 10,276,000 (`abci_query path="/store/gov/key" data=0x30`).

The SDK fork needs the companion change https://github.com/atomone-hub/atomone-sdk/pull/114, which makes the quorum getters tolerate a missing participation EMA; with this codec alone, pre-v3 state still panics on the nil EMA. Once that is released the `replace` in `go.mod` should be bumped on this branch.

---

### Author Checklist

I have...

* [x] Included the correct type prefix in the PR title
* [x] Added `!` to the type prefix if API, client, or state breaking change (not applicable: query-side decoding only; stored state and its encoding are unchanged)
* [x] Targeted the correct branch (`main`)
* [x] Provided a link to the relevant issue or specification (no issue; reproduction in the description)
* [x] Followed the guidelines for building SDK modules
* [x] Included the necessary unit and integration tests
* [x] Added a changelog entry (in the `Unreleased` section of `CHANGELOG.md`, per CONTRIBUTING)
* [x] Included comments for documenting Go code
* [x] Updated the relevant documentation or specification (not applicable)
* [x] Reviewed "Files changed" and left comments if necessary
* [ ] Confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

* [ ] confirmed the correct type prefix in the PR title
* [ ] confirmed `!` in the type prefix if API or client breaking change
* [ ] confirmed all author checklist items have been addressed
* [ ] reviewed state machine logic
* [ ] reviewed API design and naming
* [ ] reviewed documentation is accurate
* [ ] reviewed tests and test coverage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### BUG FIXES
 
 - Prevent bundling the wildcard symbol in `TxFeeExceptions` with other exceptions in `x/photon` [#352](https://github.com/atomone-hub/atomone/pull/352)
+- Decode gov params written before the v4 upgrade correctly in historical queries. Pre-v4 state uses the `atomone.gov.v1` field numbering, which the current type read one field off (v1 and v2 state) or rejected (v3 state); `TallyResult` and `Params` queries at those heights panicked or failed. The gov keeper now uses a layout-aware params codec that converts legacy bytes and pins the static quorum as the quorum range for state that predates the dynamic quorum [#366](https://github.com/atomone-hub/atomone/pull/366)
 
 ### DEPENDENCIES
 
--- a/app/keepers/keepers.go
+++ b/app/keepers/keepers.go
@@ -257,9 +257,10 @@ func NewAppKeeper(
 	govConfig := govtypes.DefaultConfig()
 	// set the MaxMetadataLen for proposals to the same value as it was pre-sdk v0.47.x
 	govConfig.MaxMetadataLen = 10200
+	govStoreService := runtime.NewKVStoreService(appKeepers.keys[govtypes.StoreKey])
 	appKeepers.GovKeeper = govkeeper.NewKeeper(
 		appCodec,
-		runtime.NewKVStoreService(appKeepers.keys[govtypes.StoreKey]),
+		govStoreService,
 		appKeepers.AccountKeeper,
 		appKeepers.BankKeeper,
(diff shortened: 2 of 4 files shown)


clockworkgr:
Closed as will-not-implement. Historical queries should be served by nodes running the software version in place at the height of the query.

